PR #21, opened 2022-07-26
Feature/swe 166 add gh page

SeanLeRoy:
The File Upload App's download page is a [Confluence page](http://confluence.corp.alleninstitute.org/display/SF/File+Upload+Application) with some out of date tutorials that largely is just used for its download links section broken up by OS. This didn't seem like the best first foot to put forward when trying to onboard people onto the app, so now it is on a similar page to the FMS File Explorer App (albeit without the tutorials for the moment) which seems like a step up. Also, no more having to remember to upgrade the download link, that aspect is done automatically. I largely just stole the patterns we used in the [FMS File Explorer App's download page](https://alleninstitute.github.io/aics-fms-file-explorer-app/).

Once this feels ready, I'll have the Confluence page just be a link to the new GH page.

[Check it out here!](https://aics-int.github.io/aics-file-upload-app/)


GabeMedrash (approved):
nice work!

kmitcham (approved):

griffinfujioka (approved):

PR #24, opened 2022-08-12
Feature/swe 236 fua max chunk

kmitcham:
This is my first commit to FUA, so I may be doing something wrong, even with my code supplied by a smart person.

### DIFF
--- a/src/renderer/services/file-management-system/index.ts
+++ b/src/renderer/services/file-management-system/index.ts
@@ -70,13 +70,8 @@ export default class FileManagementSystem {
    * This function returns the optimized number of chunks to be in flight.
   */   
   private static getInFlightChunkRequestsLimit(chunkSizeInBytes: number) {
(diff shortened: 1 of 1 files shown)


SeanLeRoy:
We are basing the FSS2 stuff off a branch called `main-fss2`, could you target this PR to that branch instead?

SeanLeRoy (approved):

kmitcham:
I see I'm allowed to commit with one review; what is the standard here?

SeanLeRoy:
> I see I'm allowed to commit with one review; what is the standard here?

This is because `main-fss2` has no branch protection rules. Just `main` does.

pgarrison (approved):
Looks good to me! If I'm understanding right, FSS2 tells the client what chunk size to use, which may be over 40MB. In that case, we'll upload a single arbitrarily large chunk at a time. What size chunks are we using?

kmitcham:
> 

Currently our chunksized is based on a fairly arbitrary constant, increased once the file is more than 10K * chunk-size.  
src/main/resources/application.properties:file.chunksize.optimized=8000000

hughes036 (approved):

PR #25, opened 2022-08-19
Remove ImageModel references

SeanLeRoy:
The concept of ImageModel has been removed from the FMS Mongo ETL and therefore also its downstream applications like the FMS File Explorer. Now that entering data for scenes, channels, timepoints, sub-images, etc. is essentially useless since it isn't shown to the user ever again, it seems to be time to remove it. Fortunately, the ImageModel feature was largely a complication so a lot of the 'upload' state is now simplified.
Once [SWE-217](https://aicsjira.corp.alleninstitute.org/browse/SWE-217) is complete, there will likely be a refactor to simplify MMS's API contract and a little more simplification can be done. For now, this seems like a big enough changeset.


kmitcham (approved):
Deleted code is the best code.

GabeMedrash (approved):

src/renderer/state/upload/selectors.ts
```diff
@@ -71,169 +66,49 @@ export const getCanUndoUpload = createSelector(
...
-    });
-    return result;
-  }
+          return annotationHasValueMap;
+        }, {} as { [annotation: string]: boolean });
+
+        return fileToAnnotationHasValueMap;
+      },
+      {} as { [file: string]: { [annotation: string]: boolean } }
+    )
```
pgarrison, line 111: This comment is not really about the changes present in this PR, just about how this function could be cleaned up in general. Feel free to ignore and merge.

My understanding of this function is that you are taking a map from filenames to metadata objects, and you are producing a map of filenames to `annotationHasValueMap`s. The structure of an `annotationHasValueMap` is a mapping from annotation names to booleans indicating whether that annotation is present for a particular file.
I think we can greatly simplify this complicated nested-reduce logic, with the following steps. (But I haven't run the code, so perhaps something about this guess is wrong). I'll use the lodash library here, since I see that is already in use in the codebase.
1. On line 97, `fileMetadata` is a regular javascript object whose keys are annotation names. Therefore, the keys are unique and only 98 each `annotation` is only present here once. This means that line 102-103 should be equivalent to `annotationHasValueMap[annotation] = !currentValueIsEmpty;`
2. Following change 1, the inner reduce (line 98) is actually performing the same computation as lodash's `_.mapValues`, and the outer reduce (line 94) is also doing a `_.mapValues`.
3. A `annotationHasValueMap` is really just a Set of annotation names (the annotations which this file has). This could allow us to use something like lodash's `_.pickBy` in place of the inner `_.mapValues`. This third change is independent of the first two and would change the contract of the function, so it may not be worth it.

Applying all three changes, this function might look something like the following:
```
function valueIsEmpty(value) {
    return isArray(value)
            ? isEmpty(value)
            : isNil(value);
)

(fileToMetadataMap): { [file: string]: Set<string> } =>
    _.mapValues(fileToMetadataMap, metadata => new Set(Object.keys(_.pickBy(metadata, valueIsEmpty))));
```

pgarrison (approved):

src/renderer/state/selection/constants.ts
```diff
@@ -2,16 +2,14 @@ import { makeConstant } from "../util";
...
+export const ADD_ROW_TO_DRAG_EVENT = makeConstant(
   BRANCH_NAME,
-  "open-sub-file-selection-modal"
+  "add-row-to-drag-event"
 );
+export const APPLY_MASS_EDIT = makeConstant(BRANCH_NAME, "apply-mass-edit");
+export const CANCEL_MASS_EDIT = makeConstant(BRANCH_NAME, "cancel-mass-edit");
 export const LOAD_FILES = makeConstant(BRANCH_NAME, "load-files");
-export const ADD_ROW_TO_DRAG_EVENT = makeConstant(
+export const OPEN_TEMPLATE_EDITOR = makeConstant(
```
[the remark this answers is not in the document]
SeanLeRoy: Ah yes bad rebase

src/renderer/state/upload/selectors.ts
SeanLeRoy: Good idea! Applied changes 1 & 2 in this commit d99b06586889a388acb1dfa385beed3b3d93944a. Didn't go forward with 3 since it seems like a neutral change but larger change.

PR #28, opened 2022-09-08
Bugfix/swe 308

hughes036:
**Purpose**:
Fix breaking changes which affect resume upload functionality (against FSS2).

**Changes**:
Updates the string values expected to be returned from FSS2, which changed in this [pr](https://github.com/aics-int/file-storage-service/pull/143/files#diff-944509ca3b449ed1afcaa574c89428b19f808e7d04c96548d104d912ae6bf388)

**How to review: where should reviewers look?**
`ChunkStatus`

**How I tested**
* Created an upload which failed on chunk 3 (inserted test code to throw error).  Then on this branch, resumed the failed upload.  Chunks resumed sending on chunk 3 based on the chunk statuses returned by FSS2 (GET /upload).
* Added a unit test covering this scenario.

Related PR:
https://github.com/aics-int/aicsfiles-python/pull/49

### DIFF
--- a/src/renderer/services/file-management-system/test/file-management-system.test.ts
+++ b/src/renderer/services/file-management-system/test/file-management-system.test.ts
@@ -13,7 +13,7 @@ import {
   MetadataManagementService,
 } from "../..";
 import { mockJob, mockWorkingUploadJob } from "../../../state/test/mocks";
-import { UploadStage, UploadStatus } from "../../file-storage-service";
+import { ChunkStatus, UploadStage, UploadStatus } from "../../file-storage-service";
 import {
   JSSJob,
   JSSJobStatus,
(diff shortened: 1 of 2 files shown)


kmitcham (approved):

GabeMedrash (approved):

src/renderer/services/file-management-system/test/file-management-system.test.ts
```diff
@@ -476,6 +476,51 @@ describe("FileManagementSystem", () => {
...
+            chunkStatuses: [ChunkStatus.COMPLETE, ChunkStatus.COMPLETE, null as any, ChunkStatus.COMPLETE, null as any, ChunkStatus.COMPLETE],
+          });
+          jss.getJob.onSecondCall().resolves(fssUpload);
+
+          // Act
+          await fms.retry("mockUploadId", noop);
+
+          // Assert
+          expect(jss.createJob.called).to.be.false;
+          const actualOffset = fileReader.read.args[0][4];
```
pgarrison, line 516: I like that we are testing at the interface between the I/O library and the application! But `args[0][4]` is uninterpretable -- this at least deserves a comment

pgarrison: I'm guessing that one of those other arguments to `fileReader.read` is the `testFilePath`? These assertions might actually be more readable if we also had an `expect(actualFilePath).to.equal(testFilePath)`, since it took me a bit of reading to figure out that that's the file we're making this assertion about

pgarrison: Ah I see I missed the boat. Here's a ping @hughes036 in case this is relevant

PR #29, opened 2022-10-03
[SWE-335] Fix slow queries for FMS.File lookup annotations 

tyler-foster:
**Context:**
[SWE-335](https://aicsjira.corp.alleninstitute.org/browse/SWE-335)
The "Optical Control ID" annotation dropdown in the File Upload App became difficult/impossible to use. The loading time after entering a FileID (at the input validation step) was either ridiculously long or never completed.

**Changes:**
* When querying for annotation values for annotations with the `Lookup` type, if the lookup is aimed at the `FMS.File` table, the query will use an "equals" filter instead of a "contains" filter. The "contains" filter is still valid for lookups on other, non-several-million-row-long tables
  * Left a TODO about how we could change this should more annotations be created that point at the `FMS.File` table
* `FMS.File` lookup annotation dropdowns no longer load the first 500 files in the table when the dropdown box is first clicked
* Some stuff got moved around by Prettier

**Testing:**
* Updated existing unit tests to pass with new `Lookup` type changes
* Manual testing
* Demo'd to one user

I'll also be making a set of tests for the `LabkeyClient` to help with changes related to SWE-279, which might capture some of the functionality present here

I originally planned to have this PR contain changes for both SWE-335 and SWE-279, but I've decided I'll make a separate PR for 279

### DIFF
--- a/src/renderer/@types/react-table-config.d.ts
+++ b/src/renderer/@types/react-table-config.d.ts
@@ -33,6 +33,8 @@ interface CustomColumnProps {
   isRequired?: boolean;
   hasSubmitBeenAttempted?: boolean;
   type?: ColumnType;
+  lookupSchema?: string;
+  lookupTable?: string;
 }
 
 // Config for `react-table` types. More info here:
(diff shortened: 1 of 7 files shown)


kmitcham (approved):

SeanLeRoy (approved):
I like that this ended up being a fairly contained fix/feature!

PR #30, opened 2022-10-12
[SWE-279] Display file names alongside ids in FMS.File Lookups

tyler-foster:
Pretty straightforward. We currently have one Lookup type annotation that references the `FMS.File` table, and users have asked to be able to see file names alongside file IDs in the dropdown selector for it. This PR adds that display in a format like:
 `File Name (File ID)`

Main change is in `labkey-client/index.ts`. In `upload/selectors.ts` there's another bit where the selected file values are updated to be in the right format for uploading, which I added a test for.


pgarrison (approved):

src/renderer/state/upload/selectors.ts
```diff
@@ -383,6 +383,23 @@ const getAnnotations = (
...
+        // We expect 'value' to be an array of strings like "<file name> (FileIdOf32CharactersAAAAAAAAAAAA)"
+        // However, we only want the fileIDs for the upload
+        if (
+          annotation.type === ColumnType.LOOKUP &&
+          annotation.lookupTable === "file"
+        ) {
+          value = value.map((fileEntry: string) => {
+            const re = RegExp(/\(([a-zA-Z0-9]{32})\)/); // Find the FileID in parentheses and capture it
+            const match = re.exec(fileEntry);
+            if (match && match.length === 2) {
```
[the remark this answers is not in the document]
kmitcham: The escaped () mean the pattern doesn't match any filenames in fms.file, although I don't know what upstream is generating the ().  
SELECT filename, created, createdby FROM fms.file WHERE filename ~ '^[a-zA-Z0-9]{32}$' 
returns zero rows in staging.

src/renderer/state/upload/test/selectors.test.ts
```diff
@@ -149,6 +150,57 @@ describe("Upload selectors", () => {
...
+          ...nonEmptyStateForInitiatingUpload.template,
+          appliedTemplate: {
+            ...mockMMSTemplate,
+            annotations: [mockFmsFileLookupAnnotation],
+          },
+        },
+        upload: getMockStateWithHistory({
+          "/path/to.dot/image.tiff": {
+            FileLookupAnnotation: [
+              "File 1 (AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA)",
```
kmitcham, line 166: in response to Phillips point, maybe a test where the string is
"XAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAX (AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA)"
and make sure you get the Second value back?
For that matter, you might want to test 
"File 1 (AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA)"  - short a character
"File 1 AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" missing ()
"File 1 (AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA)" extra character

kmitcham (approved):

src/renderer/services/labkey-client/index.ts
```diff
@@ -177,16 +179,12 @@ export default class LabkeyClient extends HttpCacheClient {
...
-      `query.sort=${column}`,
-      `query.maxRows=100`,
-    ];
+    const additionalQueries = [`query.sort=${column}`, `query.maxRows=100`];
     if (!isEmpty(searchString)) {
       // The fms.file table is so large that running "contains" queries on it will take forever, so use "eq" instead
       /*
        TODO:
-        In a more perfect world, annotations that reference other files (specifically, files fitting certain annotation
+        In a more perfect world, annotations that reference other files (specifically files fitting certain annotation
```
hughes036, line 187: nit i think it reads better with the ','

hughes036 (approved):

src/renderer/state/upload/test/selectors.test.ts
tyler-foster: Mm yeah, those are good ideas. I'll add some with varying file name lengths too.

src/renderer/services/labkey-client/index.ts
tyler-foster: Obviously it doesn't matter very much, but I think I removed the comma here because I don't like repeating sentence structures one after the other, and I thought the stuff in parentheses read that way.
e.g.
`Beginning 1, main point 1. Beginning 2, main point 2`

src/renderer/state/upload/selectors.ts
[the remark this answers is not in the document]
tyler-foster: > Maybe a more robust option would be to take the last 34 characters of `fileEntry`, check that they match the pattern, and then strip off the parentheses from that?

Ah yeah, that's essentially what it's supposed to be doing but it looks like I missed the bits necessary to make sure I'm only matching on the end of `fileEntry`

src/renderer/services/labkey-client/index.ts
```diff
@@ -177,16 +179,12 @@ export default class LabkeyClient extends HttpCacheClient {
     column: string,
     searchString = ""
   ): Promise<string[]> {
-    const additionalQueries = [
-      `query.columns=${column}`,
```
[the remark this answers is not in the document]
tyler-foster: True, I wouldn't expect it to make a huge difference but I can definitely slim that down

src/renderer/state/upload/test/selectors.test.ts
```diff
@@ -163,8 +163,15 @@ describe("Upload selectors", () => {
...
-              "File 2 (BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB)",
+              // Over-testing a bit to make sure our regex works. We really only want to upload FileIDs
+              "File 1 (File1AAAAAAAAAAAAAAAAAAAAAAAAAAA)",
+              "File 2 (File2AAAAAAAAAAAAAAAAAAAAAAAAAAA)",
+              "File 3 with a considerably longer file name (File3AAAAAAAAAAAAAAAAAAAAAAAAAAA)",
+              "File 4",
+              "File 5 parens do not contain ID (A)",
+              "File 6 no paren - File6AAAAAAAAAAAAAAAAAAAAAAAAAAA",
+              "File 7 missing paren (File7AAAAAAAAAAAAAAAAAAAAAAAAAAA",
+              "File 8 extra parens (File8AAAAAAAAAAAAAAAAAAAAAAAAAAA) (File8AAAAAAAAAAAAAAAAAAAAAAAAAAA)",
```
kmitcham, line 174: nit: I'd use
"File 8 extra parens (File8AAAAAAAAAAAAAAAAAAAAAAAAABB) (File8AAAAAAAAAAAAAAAAAAAAAAAAAAA)",
to be sure it was grabbing the right value.

PR #34, opened 2022-10-27
[SWE-450] Fix issue where uploads can't be edited if they contain unexposed annotations

tyler-foster:
#### Context
This PR address an error that was being thrown when users attempted to "view" a file with the "MXS Processing Failed" annotation. The logic for mapping out the annotation info for a given file was only looking for annotations to expose to the FUA. However, this became inappropriate with the addition of these new hidden annotations, and was already redundant given that the FUA later only displays the exposed annotations

#### Testing
I added one `logics` test to ensure that we can make it to the file viewing/editing page regardless of what annotations are attached to the file being viewed. It fails while using the main branch's code and succeeds after my changes.
I also manually tested one file I had available to me that had failed during MXS processing - I've confirmed that I can open and edit its metadata now

### DIFF
--- a/src/renderer/state/test/mocks.ts
+++ b/src/renderer/state/test/mocks.ts
@@ -101,7 +101,7 @@ export const mockNotesAnnotation: Annotation = {
 
 const mockUnusableStructureAnnotation: Annotation = {
   ...mockAuditInfo,
-  annotationId: 3,
+  annotationId: 4,
   annotationTypeId: 1,
   description: "Other information",
   exposeToFileUploadApp: false,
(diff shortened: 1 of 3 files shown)


kmitcham (approved):

src/renderer/state/test/mocks.ts
```diff
@@ -421,6 +421,45 @@ export const mockSuccessfulUploadJob: JSSJob = {
...
+  created: new Date(),
+  currentStage: "Completed",
+  jobId: "123434234",
+  jobName: "mockJob1",
+  modified: new Date(),
+  serviceFields: {
+    files: [
+      {
+        customMetadata: {
+          annotations: [
```
hughes036, line 434: I am not familiar with the concept of exposed vs unexposed annotations - what makes these 'unexposed'?

hughes036 (approved):

src/renderer/state/test/mocks.ts
tyler-foster: [Some annotations are generated ](http://aics.corp.alleninstitute.org/labkey/query/AICS/executeQuery.view?schemaName=filemetadata&query.queryName=Annotation&query.ExposeToFileUploadApp~eq=false)in places other than the FUA, such as in the LabKey plate custom data tab or via MXS, so they're marked to not be "exposed" to the FUA so people don't accidentally use them there. They still get used later down the line in other processing steps / the Explorer app though.
e.g. You don't want people uploading files via the app while marking the "MXS Processing Failed" annotation as true or false -  it wouldn't make sense for the user to declare that either way

PR #35, opened 2022-11-03
FSS2 Performance: Make FSS and JSS requests concurrently

pgarrison:
**The issue**
FUA against FSS2 takes [about twice as long](http://confluence.corp.alleninstitute.org/display/SF/FSS2+Benchmark+Testing) as with FSS-current. This change doesn't have a ticket because it's little code and I found it accidentally.

**The change**
Previously, for each chunk in an upload FUA would wait for the update to the partial MD5 in JSS to complete before sending the chunk to FSS. This change would execute both requests at once, hopefully for improved performance.

**Testing**
`yarn test`

But the real test is whether it improves performance. I haven't done a performance test. If I do, not sure what I should name the release.

### DIFF
--- a/src/renderer/services/file-management-system/index.ts
+++ b/src/renderer/services/file-management-system/index.ts
@@ -506,7 +506,7 @@ export default class FileManagementSystem {
     // Handles submitting chunks to FSS, and updating progress
     const uploadChunk = async (chunk: Uint8Array, chunkNumber: number, md5ThusFar: string): Promise<void> => {
       chunksInFlight++;
-      await this.jss.updateJob(uploadId, {
+      const jssPromise = this.jss.updateJob(uploadId, {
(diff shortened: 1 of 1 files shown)


kmitcham (approved):
Every little bit helps

SeanLeRoy:
This happening sequentially was intentional to avoid FSS getting ahead of JSS in the event the FSS requests succeeds but the JSS request does not. As you have this written currently there is a case we'd be unable to resume or produce the wrong MD5.

SeanLeRoy (changes requested):
This happening sequentially was intentional to avoid FSS getting ahead of JSS in the event the FSS requests succeeds but the JSS request does not. As you have this written currently there is a case we'd be unable to resume or produce the wrong MD5.

PR #42, opened 2022-11-21
[SWE-565] Fix linux, macOS builds on main

tyler-foster:
This PR aims to address a couple of weird things that have been going on with the GH actions release recently.

- The first thing was that the macOS build was failing due to GH's `mac-latest` upgrading to macOS 12, which was reasonable. This PR addresses the issue by simply upgrading the `electron-builder` package version to `v23`
- The second was that our linux build was failing due to a borked Enzyme function, which is uhh... still unsolved. `TypeError: X is not a function`. What does this mean? We aren't sure. Only one test was failing this way, and the issue isn't happening on the FSS2 trunk, so Sean and I agreed that it should be alright to just comment it out for now, since we're aiming to merge the two trunks soon-ish anyways. If the problem persists after that, well... that's a work item.
- This PR also fixes another test that was throwing a similar error, but this one was fixed by just upgrading `enzyme` and changing one line of code

Ignore the rapid back-to-back snapshot tags. I got kinda silly on a different branch


pgarrison:
`main-FSS2` has `"@types/enzyme": "~3.10.11",`, while it's `3.9.0` here. Perhaps that is an easier upgrade than moving to `enzyme-adapter-react-17`. No clue if it will help though

pgarrison (approved):
In the TemplateSearch component, I see the type specification `onSelect: (selectedTemplateId: number) => void;`, but 1 in `() => 1` is not void. Perhaps `() => {}` will work?

pgarrison:
Or perhaps the `store` from `createMockReduxStore` is missing something? I guess you could narrow down the location of the bug by trying to mount a `Provider` without a `TemplateSearch`?

griffinfujioka (approved):

tyler-foster:
Merging now, but thank you @pgarrison for the extra ideas for investigation

PR #45, opened 2022-11-30
Bugfix/swe 571

hughes036:
I believe this is all that is necessary on the FUA side.

When metadata is initially attached to a freshly upload file, FUA will use `PATCH /filemetadata/`

For subsequent edits, it will _continue_ to use `PUT /filemetadata/`

No unit tests needed updating for this change.  I plan to rely on manual testing, when we have the new endpoint deployed on stg.

**NEEDED BEFORE MERGE:** 
Manual testing


package.json
```diff
@@ -53,10 +53,10 @@
     "@aics/frontend-insights": "~0.2.3",
     "@aics/frontend-insights-plugin-amplitude-node": "~0.2.3",
     "@ant-design/icons": "~4.7.0",
-    "@types/spark-md5": "^3.0.2",
     "@types/react-beautiful-dnd": "~13.1.2",
     "@types/react-virtualized-auto-sizer": "~1.0.0",
     "@types/react-window": "~1.8.3",
+    "@types/spark-md5": "^3.0.2",
```
SeanLeRoy, line 59: Oh I didn't notice this when this was added, but this isn't using the same version restrictions as the others `^` vs `~`. Could you switch this to be `~3.0.2` along with `spark-md5` below on line 83? 
`^` means it will auto-upgrade to patch & minor versions of the same major version.
`~` means it will auto-upgrade to patch version of the same major and minor version (more restrictive)

SeanLeRoy (commented):
Nice 👍

kmitcham (approved):

src/renderer/state/route/logics.ts
```diff
@@ -289,38 +296,37 @@ const viewUploadsLogic = createLogic({
...
-              originalPath: labkeyFileMetadata.localFilePath as string,
-              fileType: labkeyFileMetadata.fileType,
-            },
-          };
-        })
-      );
-      const uploadsToView = convertUploadRequestsToUploadStateBranch(
-        [...requests, ...fileIdsAsFiles],
-        state
+      const metadataForFiles = await Promise.all(fileIds.map(async (fileId) => await Promise.all([
```
pgarrison, line 299: I think this is equivalent to just
```
      const metadataForFiles = await Promise.all(fileIds.map(fileId => Promise.all([
      ...
```
Because `return await foo();` is the same as `return foo();` for an async function `foo` (async functions return promises, and `return await x;` means "return a promise that resolves to whatever the promise `x` resolves to)

pgarrison (approved):

PR #46, opened 2022-12-06
[SWE-581] Reset upload once confirmed

SeanLeRoy:
Upload data was not being cleared from the state after the upload was initiated and the user was moved to the "My Uploads" page. This led to some confusing behavior where to do another upload the user had to click on "Current Upload" and click the "Clear" button discarding the upload data themselves then coming back and starting a new upload from the "My Uploads" page again.

### DIFF
--- a/src/renderer/state/upload/logics.ts
+++ b/src/renderer/state/upload/logics.ts
@@ -220,7 +220,10 @@ const initiateUploadLogic = createLogic({
       return;
     }
 
-    dispatch(initiateUploadSucceeded(action.payload));
+    dispatch(batchActions([
+      initiateUploadSucceeded(action.payload),
+      resetUpload()
+    ]))
 
     const uploadTasks = uploads.map((upload) => async () => {
       try {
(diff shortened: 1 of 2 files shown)


src/renderer/state/upload/logics.ts
```diff
@@ -220,7 +220,10 @@ const initiateUploadLogic = createLogic({
       return;
     }
 
-    dispatch(initiateUploadSucceeded(action.payload));
+    dispatch(batchActions([
+      initiateUploadSucceeded(action.payload),
+      resetUpload()
```
hughes036, line 225: Any straight forward way to unit test this?

hughes036 (approved):

kmitcham (approved):

tyler-foster (approved):

src/renderer/state/upload/logics.ts
SeanLeRoy: Good idea, added

PR #47, opened 2022-12-19
[SWE-387] Move partial md5 into fss

SeanLeRoy:
This moves the partial MD5 persistence necessary to be able to resume an upload at any chunk from JSS to FSS which reduces the HTTP requests we have to make and moves chunk information into one place.

**Testing**
Started, stopped, and resumed uploads against staging

### DIFF
--- a/src/renderer/services/file-management-system/index.ts
+++ b/src/renderer/services/file-management-system/index.ts
@@ -462,7 +462,7 @@ export default class FileManagementSystem {
             throw new Error('No partial MD5 for chunk ' + lastChunkNumber);
           }
           const md5 = Md5Hasher.deserialize(partiallyCalculatedMd5).digest();
-          await this.fss.retryFinalize(fssUploadId, md5)
+          await this.fss.finalize(fssUploadId, md5)
         } else if (fssStatus.status === UploadStatus.WORKING) {
           const { originalPath } = upload.serviceFields.files[0].file;
           const { size: fileSize } = await fs.promises.stat(originalPath);
--- a/src/renderer/services/file-management-system/index.ts
+++ b/src/renderer/services/file-management-system/index.ts
@@ -6,7 +6,6 @@ import * as uuid from "uuid";
 
 import { extensionToFileTypeMap, FileType } from "../../util";
 import FileStorageService, {
-  ChunkStatus,
   FSSUpload,
   UploadStatus,
   UploadStatusResponse,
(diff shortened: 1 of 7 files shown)


kmitcham (approved):

src/renderer/services/file-management-system/index.ts
```diff
@@ -447,25 +442,31 @@ export default class FileManagementSystem {
         // If FSS is still available to continue receiving chunks of this upload
         // simply continue sending the chunks
         let lastChunkNumber = fssStatus.chunkStatuses.findIndex(
-          (status) => status !== ChunkStatus.COMPLETE
+          (status) => status !== UploadStatus.COMPLETE
```
hughes036, line 445: Is this change correct?  They likely have the same value, but ChunkStatus seems like the right place to store it.

src/renderer/services/file-management-system/index.ts
```diff
@@ -447,25 +442,31 @@ export default class FileManagementSystem {
...
+        if (lastChunkNumber > 0) {
+          const chunkResponse = await this.fss.getChunkInfo(fssUploadId, lastChunkNumber);
+          partiallyCalculatedMd5 = chunkResponse.cumulativeMD5;
+          if (!partiallyCalculatedMd5){
+            throw new Error('No partial MD5 for chunk ' + lastChunkNumber);
+          }
         }
 
-        if (fssStatus.status === UploadStatus.RETRY ) {
+        if (fssStatus.status === UploadStatus.RETRY) {
```
hughes036, line 460: Does this condition need to be concerned about `partiallyCalculatedMd5 ` ?  If not It looks like this condition could come first, and that would simplify the routine.

pgarrison (approved):

src/renderer/services/file-management-system/index.ts
SeanLeRoy: If they are always the same enum values then it makes sense to me to combine them

src/renderer/services/file-management-system/index.ts
SeanLeRoy: Yeah it uses it in the retryFinalize call, is that unnecessary to give to FSS on retry?

hughes036: Yeah `retryFinalize` doesnt take and MD5.

But also remember, after https://github.com/aics-int/file-storage-service/pull/210 , RETRY will just mean for the client to call `/finalize` again.  When called in a retry scenerio, the MD5 field can be blank.  If it is not blank, it must have the same value as when `/finalize` was called the first time.

src/renderer/services/file-management-system/index.ts
hughes036: Yeah i agree.  I just think the name of the enum shouldn't be so specific to Upload then.

hughes036 (approved):

PR #49, opened 2023-01-05
[SWE-491] Retry chunk when possible

SeanLeRoy:
Previously chunks were tolerant to timeouts due to the loadbalancer, this adds a feature building off of that to auto-retry chunks that are reported to of status "RETRY" by FSS meaning they encountered some issue but are recoverable.

**Testing**
Unit Test + Manually tested by modifying the return from FSS locally

### DIFF
--- a/src/renderer/services/file-storage-service/index.ts
+++ b/src/renderer/services/file-storage-service/index.ts
@@ -19,10 +19,10 @@ export interface FSSUpload extends JSSJob {
  */
 export enum UploadStatus {
   COMPLETE = "COMPLETE",
-  WORKING = "WORKING",                  //Upload is in progress and accepting chunks.
-  INACTIVE = "INACTIVE",                //Upload was either cancelled, expired, or failed.
-  RETRY = "RETRY",                      //Upload experienced a recoverable error, and can resume when /upload/retry is called.
-  POST_PROCESSING = "POST_PROCESSING"   //Chunks were all recieved, /finalize was called, and post upload processing is happening.  
+  WORKING = "WORKING",                  // Upload is in progress and accepting chunks.
+  INACTIVE = "INACTIVE",                // Upload was either cancelled, expired, or failed.
+  RETRY = "RETRY",                      // Upload experienced a recoverable error, and can resume when /upload/retry is called.
+  POST_PROCESSING = "POST_PROCESSING"   // Chunks were all recieved, /finalize was called, and post upload processing is happening.  
 }
 
 // RESPONSE TYPES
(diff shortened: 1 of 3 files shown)


src/renderer/services/file-storage-service/test/file-storage-service.test.ts
```diff
@@ -60,6 +61,49 @@ describe("FileStorageService", () => {
...
+    class AxiosError extends Error {
+      public response: any;
+
+      constructor(response: any) {
+        super("test error");
+        this.response = response;
+      }
+    }
+
+    it("retries chunk when error occurs and server status indicates RETRY", async () => {
```
kmitcham, line 73: No test of the error when the number of retries is exceeded.  Is that intentional?

kmitcham (approved):

griffinfujioka (approved):

src/renderer/services/file-storage-service/index.ts
```diff
@@ -114,25 +114,68 @@ export default class FileStorageService extends HttpCacheClient {
...
+        },
+      }
+      )
+    );
+    try {
+      await attemptRequest();
+    } catch (error) {
+      // Re-throw error if status can't be determine (like when the error doesn't have to do with the server/HTTP request
+      // or if the status is between acceptable ranges indicating another problem is afoot
+      if (!error.response || !error.response.status || (error.response.status >= 200 && error.response.status < 300)) {
```
hughes036, line 144: Is it possible for an error to have a status in the 200 range?  I think that would be a success.

src/renderer/services/file-storage-service/index.ts
```diff
@@ -114,25 +114,68 @@ export default class FileStorageService extends HttpCacheClient {
...
+        throw error;
+      }
+
+      // Currently there is a bug in FSS that causes the upload chunk endpoint
```
hughes036, line 148: > Currently there is a bug in FSS that causes the upload chunk endpoint

Currently there _are infrastructure performance bottlenecks_ that causes the upload chunk endpoint 

?

hughes036 (approved):

src/renderer/services/file-storage-service/index.ts
SeanLeRoy: Yeah technically it cant have issues decoding the response I believe, it is a rare case though

src/renderer/services/file-storage-service/test/file-storage-service.test.ts
SeanLeRoy: I didn't have a good way to test this that didn't slow the test suite down significantly

PR #50, opened 2023-01-17
Use hash-wasm instead of spark-md5

SeanLeRoy:
The hope is that using `hash-wasm` will get us farther away from being IO bound since this package relies on web assembly instead of `spark-md5` which seems to have proven itself to force uploads to be CPU bound.


kmitcham (approved):
I am still pleased and impressed you got the serializable MD5 stuff.

griffinfujioka (approved):

src/renderer/services/file-management-system/Md5Hasher.ts
```diff
@@ -1,46 +1,70 @@
...
      */
-     public static deserialize (serialized_md5: string) : Md5Hasher{ 
-        const hasher = new ArrayBuffer();
-        hasher.setState(JSON.parse(serialized_md5));
+     public static async deserialize(serializedMd5: string): Promise<Md5Hasher>{
+        // The load() function necessary to rehydrate the Md5Hasher state requires a Uint8Array,
+        // but in order to persist the Uint8Array it had to be converted to string.
+        // This intends to take an array-like string and convert it back into an array
+        // that a Uint8Array can be created from.
+        const serializedMd5AsArray = serializedMd5.split(Md5Hasher.ARRAY_DELIMITER).map(v => +v);
```
hughes036, line 37: This looks like it maps each delimited string v using `(v => +v)`.  Im a bit confused by that, can you explain?

hughes036 (approved):

src/renderer/services/file-management-system/Md5Hasher.ts
SeanLeRoy: The `+` converts the number-like string into an integer

PR #53, opened 2023-01-25
[SWE-376] Remove CSC envs for Windows release in release.yml

tyler-foster:
This _should_ remove the code-signing step in the Windows build process, if I understand correctly


tyler-foster:
Insta-merging per my own discretion...

PR #54, opened 2023-03-07
[UR-93] Highlight annotations that trigger post-processing

pgarrison:
Make users more aware of `Is Optical Control`, `Optical Control ID`, and `Well` when creating a new template. The alert added here is brief because each annotation has a longer description that is shown both when creating a template and (in a tooltip) when doing an upload with a metadata template.

### Testing
Manual & `yarn test`.
Screenshot 2023-03-06 at 3 57 57 PM
Screenshot 2023-03-06 at 3 58 15 PM

### DIFF
--- a/src/renderer/containers/TemplateEditorModal/index.tsx
+++ b/src/renderer/containers/TemplateEditorModal/index.tsx
@@ -28,7 +28,7 @@ import {
   getTemplateEditorVisible,
 } from "../../state/feedback/selectors";
 import { getAnnotationsWithAnnotationOptions } from "../../state/metadata/selectors";
-import { getShowTemplateHint } from "../../state/setting/selectors";
+import { getShowTemplateHint, getShowAddAnnotationHint } from "../../state/setting/selectors";
 import {
   addExistingAnnotation,
   addExistingTemplate,
@@ -54,6 +54,9 @@ const styles = require("./styles.pcss");
 
 const { Search } = Input;
 
+const ADD_ANNOTATION_DESCRIPTION = `Certain annotations trigger automated processes for new uploads.
+Optical Control ID and Is Optical Control are used to submit images for automatic alignment.
+Well is used to automatically add metadata from Labkey.`
 const COLUMN_TEMPLATE_DESCRIPTION = `A ${SCHEMA_SYNONYM} defines a group of annotations to associate with files.
 When applied to a batch of files to upload, the annotations associated with that template
 will be added as additional columns to fill out for each file. They can be shared and discovered by anyone.`;
@@ -108,6 +111,7 @@ function TemplateEditorModal(props: Props) {
   const dispatch = useDispatch();
   const template = useSelector(getTemplateDraft);
   const originalTemplate = useSelector(getOriginalTemplate);
+  const showAnnotationHint = useSelector(getShowAddAnnotationHint);
   const showTemplateHint = useSelector(getShowTemplateHint);
   const allAnnotations = useSelector(getAnnotationsWithAnnotationOptions);
   const requestsInProgress = useSelector(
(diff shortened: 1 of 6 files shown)


src/renderer/containers/TemplateEditorModal/index.tsx
```diff
@@ -54,6 +54,9 @@ const styles = require("./styles.pcss");
 
 const { Search } = Input;
 
+const ADD_ANNOTATION_DESCRIPTION = `Certain annotations trigger automated processes for new uploads.
+Optical Control ID and Is Optical Control are used to submit images for automatic alignment.
```
hughes036, line 58: Do users understand how they can find the aligned images?

src/renderer/state/setting/test/logics.test.ts
```diff
@@ -27,6 +27,7 @@ import {
 import { updateSettingsLogic } from "../logics";
 import {
   getLimsHost,
+  getShowAddAnnotationHint,
```
hughes036, line 30: looks like this can be removed (according to the build alert)?

hughes036 (approved):

src/renderer/state/setting/test/logics.test.ts
pgarrison: Ah I had added the wrong test -- now it should be used

src/renderer/containers/TemplateEditorModal/index.tsx
pgarrison: I think we don't have enough aligned images to really answer that question

src/renderer/containers/TemplateEditorModal/index.tsx
```diff
@@ -54,6 +54,9 @@ const styles = require("./styles.pcss");
...
+Well is used to automatically add metadata from Labkey.`
```
SeanLeRoy, line 59: I'm now wondering if "Well" should just be excluded altogether since it is included by default and we really don't want them adding "Well" to templates.

src/renderer/state/types.ts
```diff
@@ -287,6 +287,7 @@ export interface EnabledNotifications {
 
 export interface SettingStateBranch extends LimsUrl {
   metadataColumns: string[];
+  showAddAnnotationHint: boolean;
```
SeanLeRoy, line 290: There doesn't appear to be a way to mark this as false, so it seems it is perpetually true. Regardless, the way the `showUploadHint` and `showTemplateHint` are toggled is within the settings modal and I think this "add annotation" hint should be considered a template hint and controlled by that variable rather than as an additional more niche setting

SeanLeRoy (changes requested):

src/renderer/containers/TemplateEditorModal/index.tsx
hughes036: Do they get annotated in a way that links it to the original?  It seems like the ideal thing would be to give the original an annotation which links it to the derived image(s).

src/renderer/containers/TemplateEditorModal/index.tsx
pgarrison: How do you know "Well" is a default? It appears in the list of annotations to add to a new template, and I see templates without Well (e.g., AICS CofA Images). Only Notes and Plate Barcode are included in all of the current templates

src/renderer/state/types.ts
pgarrison: I didn't really understand what the settings were for -- fixed

src/renderer/containers/TemplateEditorModal/index.tsx
SeanLeRoy: Well isn't shown until a Plate Barcode is selected so by that reasoning it is always included since Plate Barcode is always included.

pgarrison: Ohhh, great

PR #55, opened 2023-03-07
Bugfix/swe 784

hughes036:
**Situation**:  Hailey demonstrated that the app would crash while uploading an especially large (+4tb) file.

**Repro Steps**: I was able to reproduce the behavior by hardcoding a large chunksize (simulating the chunksize generated for a large file).

**Steps to fix**: Next, I hardcoded chunksInFlight to various numbers, and confirmed that by keeping below a threshold, that prevented the app from crashing.

**Intervention**: Dynamically size maxChunksInFlight to keep the overall bytesInFlight below the threshold.

**Testing**: Manually observing behavior, unit testing that the algorithm produces the right values.

**TODO**: This is a bugfix (yet to be verified by a real +4tb file), but we would like to understand the mechanism better and probably implement a better solution. 
 We don't have a great way to measure the actual memory footprint, since V8 crashes, preventing a snapshot from saving.  The performance devtool also displays a heapSpace value that seems suspect, since it never goes above 120MB when the chunks in flight would sum to much higher than that.


src/renderer/services/file-management-system/index.ts
```diff
@@ -481,6 +484,16 @@ export default class FileManagementSystem {
...
+  public static chunksInFlightMaxForChunkSize(chunkSize: number){
+    for( let chunksInFlight = FileManagementSystem.CHUNKS_CEILING_IN_FLIGHT_REQUEST_CEILING_DEFAULT; chunksInFlight > 0; chunksInFlight-- ){
+      const bytesInFlight = chunksInFlight * chunkSize;
+      if(bytesInFlight <= FileManagementSystem.BYTES_IN_FLIGHT_CEILING){
+        return chunksInFlight;
+      }
+    }
+    return 1;//Should never reach here.
+  }
+
```
pgarrison, line 496: Instead of this function, perhaps `Math.floor(BYTES_IN_FLIGHT_CEILING/chunkSize)`?

src/renderer/services/file-management-system/index.ts
```diff
@@ -511,7 +524,11 @@ export default class FileManagementSystem {
     
     // For rate throttling how many chunks are sent in parallel
     let chunksInFlight = 0;
-    const chunksInFlightLimit = FileManagementSystem.CHUNKS_CEILING_INFLIGHT_REQUEST_CEILING;
+
+    const chunksInFlightLimit = FileManagementSystem.chunksInFlightMaxForChunkSize(chunkSize);
+    //TODO remove after we observe bugfix effacacy in the field -chrishu 3/7/23
+    console.log("chunksInFlightLimit: " + chunksInFlightLimit)
+    console.log("chunkSize: " + chunkSize)
```
pgarrison, line 531: Should we release this as a snapshot from this branch and then remove the log statements before merging this PR?

src/renderer/services/file-management-system/index.ts
hughes036: wish I had though of that . .

hughes036: Though I think we still want to constrain it at 20 for small chunkSizes (for now at least).

pgarrison:
Some quick & dirty performance testing uploading 10GB files to staging
* From commit `62e76a31` (50MB chunk size, 20 chunks in flight, should be the same as previous tests): 410 seconds, 24MB/s
* From commit `62e76a31` but with chunk size hard-coded to 500MB (4 chunks in flight): 571 seconds, 17.5MB/s

hughes036:
> Some quick & dirty performance testing uploading 10GB files to staging
> 
> * From commit `62e76a31` (50MB chunk size, 20 chunks in flight, should be the same as previous tests): 410 seconds, 24MB/s
> * From commit `62e76a31` but with chunk size hard-coded to 500MB (4 chunks in flight): 571 seconds, 17.5MB/s
> * From commit `62e76a31` but with chunk size hard-coded to 500MB and chunks in flight hard-coded to 5: 455 seconds, 22MB/s

I think the upshot (for this sample) is: 

More chunks in flight produces greater throughput, even though the overall bytes in flight is lower.

Right?

src/renderer/services/file-management-system/index.ts
```diff
@@ -48,7 +48,10 @@ export default class FileManagementSystem {
   private readonly jss: JobStatusService;
   private readonly mms: MetadataManagementService;
 
-  private static readonly CHUNKS_CEILING_INFLIGHT_REQUEST_CEILING = 20; //ceiling on concurrent chunk requests (even if more can fit in memory)
+  private static readonly CHUNKS_IN_FLIGHT_REQUEST_CEILING_DEFAULT = 20; //ceiling on concurrent chunk requests (even if more can fit in memory)
+  private static readonly LARGEST_CUNK_SIZE = 432000000 //Max chunk size tested, from a bug report.  Generated for a 4+TB file SWE-784
```
SeanLeRoy, line 52:
```suggestion
  private static readonly LARGEST_CHUNK_SIZE = 432000000 //Max chunk size tested, from a bug report.  Generated for a 4+TB file SWE-784
```

src/renderer/services/file-management-system/index.ts
```diff
@@ -48,7 +48,10 @@ export default class FileManagementSystem {
...
+  private static readonly LARGEST_SUCCESSFUL_CHUNKS_IN_FLIGHT = 5; //Largest value used (in combination with LARGEST_CUNK_SIZE) in a succesfull test SWE-784
```
SeanLeRoy, line 53:
```suggestion
  private static readonly LARGEST_SUCCESSFUL_CHUNKS_IN_FLIGHT = 5; //Largest value used (in combination with LARGEST_CHUNK_SIZE) in a successful test SWE-784
```

src/renderer/services/file-management-system/test/file-management-system.test.ts
```diff
@@ -576,4 +576,14 @@ describe("FileManagementSystem", () => {
       await expect(fms.cancel(mockUploadId)).rejectedWith(Error);
     });
   });
+
+  describe("Dynamically set chunksInFlight according to chunkSize", () => {
```
SeanLeRoy, line 580: This is kinda a nit but it breaks the pattern the other tests in this repo have where describe is used to establish the method being tested and then the inner "it"s are the tests. Like so:
```
describe("chunksInFlightMaxForChunkSize", () => {
    [
         { "chunkSize": 50000000, "expectedChunksInFlight": 20 },
         ....(the other test cases)
     ].forEach(({ chunkSize, expectedChunksInFlight }) => {
          it(`Dynamically sets chunksInFlight to ${expectedOutput} for chunkSize ${input}`, () => {
                  expect(FileManagementSystem.chunksInFlightMaxForChunkSize(chunkSize)).to.equal(expectedChunksInFlight);
           })
     })
})
```

kmitcham (approved):
I had a similar chunksize issue with the java testing client.

SeanLeRoy:
What is the status on this PR? Is the current state ready for review or is it WIP?

PR #56, opened 2023-03-27
Bugfix/swe 784 2

hughes036:
Situation: Hailey demonstrated that the app would crash while uploading an especially large (+4tb) file.

Repro Steps: I was able to reproduce the behavior by hardcoding a large chunksize (simulating the chunksize generated for a large file).

Intervention: Dynamically size maxChunksInFlight to keep the overall bytesInFlight below the threshold.

- Dynamically compute `chunksInFlight` based on chunkSize, with a ceiling of 10 for small chunks.  The aim is to keep the summed "bytes in flight" under 2GB, while also not needlessly queuing huge numbers of small requests (chromium can only have 6 pending requests).
- Before a new chunk request is made, check the `external` memory metric.  If it is above 2GB, manually call garbage collection.  Based on my observations and reading, a 'race condition' is possible when allocating new memory (reading a chunk) and garbage collection, which can result in a consumption spike before necessary memory is freed by GC.

Info about the `external` metric: https://nodejs.org/api/process.html#processmemoryusage

Testing: Manually observing behavior, debug output.

TODO: This is a bugfix (is being verified by Hayley on a +4tb file), but we would like to understand the mechanism better and probably implement a better solution: [SWE-822](https://aicsjira.corp.alleninstitute.org/browse/SWE-822)


kmitcham (approved):

src/main/index.ts
```diff
@@ -96,6 +96,8 @@ function createMainWindow() {
 
 // This is a temporary fix for this issue: https://github.com/electron/electron/issues/23664
 app.commandLine.appendSwitch("disable-features", "OutOfBlinkCors");
+// This is to enable maula garbage collection SWE-784
```
SeanLeRoy, line 99:
```suggestion
// This is to enable manual garbage collection SWE-784
```

src/renderer/services/file-management-system/index.ts
```diff
@@ -495,6 +497,10 @@ export default class FileManagementSystem {
   }): Promise<void> {
     const { fssUploadId, source, chunkSize, user, onProgress, initialChunkNumber = 0, partiallyCalculatedMd5 } = config;
     let chunkNumber = initialChunkNumber;
+    const dynamicChunksInFlight = Math.floor(FileManagementSystem.EXTERNAL_BYTES_USED_CEILING/chunkSize);
+    const chunksInFlightLimit = Math.min(dynamicChunksInFlight, 10);
+    console.log("chunksInFlightLimit: " + chunksInFlightLimit);
+    console.log("GC enabled: " + new Boolean(global.gc !== undefined));
```
SeanLeRoy, line 503:
```suggestion
    console.log("GC enabled: " + (global.gc !== undefined));
```

SeanLeRoy (commented):
Is this working? If not, maybe this is better as a draft PR until ready for review? If yes, awesome!

hughes036:
> Is this working? If not, maybe this is better as a draft PR until ready for review? If yes, awesome!

It "works", it is still being vetted on a 4tb file, on Windows.  I think this is as far as it is worth going down this road.  Further efforts should be spent on [SWE-822](https://aicsjira.corp.alleninstitute.org/browse/SWE-822)

src/renderer/services/file-management-system/index.ts
```diff
@@ -540,10 +545,19 @@ export default class FileManagementSystem {
      * side effect of populating uploadChunkPromises.
      */
     const onChunkRead = async (chunk:Uint8Array, md5ThusFar: string): Promise<void> => {
-      // Throttle how many chunks will be uploaded in parallel
+      // Throttle how many chunks will be loaded into memory
       while (chunksInFlight >= chunksInFlightLimit) {
+        console.log(process.memoryUsage());
```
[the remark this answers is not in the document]
hughes036: I'm planning to remove it before merging, right now its just here for me and @pgarrison

PR #57, opened 2023-04-06
revise restart code for cases when JSS status is failed, but FSS is w…

kmitcham:
This needs at least more testing, and unit test coverage before merging, but getting this out in case others want to pick it up

### DIFF
--- a/src/renderer/services/file-management-system/index.ts
+++ b/src/renderer/services/file-management-system/index.ts
@@ -128,7 +128,6 @@ export default class FileManagementSystem {
         fileType,
         fileSize,
       );
-
       // Update parent job with upload job created by FSS
       // for tracking in the event of a retry
       await this.jss.updateJob(upload.jobId, {
@@ -137,15 +136,26 @@ export default class FileManagementSystem {
           lastModifiedInMS: fileLastModifiedInMs,
         },
       });
-
-      // Wait for upload
-      await this.uploadInChunks({
-        fssUploadId: registration.uploadId,
-        source: source,
-        chunkSize: registration.chunkSize,
-        user: upload.user,
-        onProgress: (bytesUploaded) => onProgress({ bytesUploaded, totalBytes: fileSize })
-      });
+      if (registration.chunkStatuses && registration.chunkStatuses[0]){
+        // update the maongodb fua.upload job with the fss uplaod_id field
+        const onRetryProgress = (
+            uploadId: string,
+            progress: UploadProgressInfo
+        ) => {
+          onProgress(progress)
+        };
+        // create the FUA jss uplaod id and info here
+        await this.retry(upload.jobId, onRetryProgress);
+      } else {
+        // Wait for upload
+        await this.uploadInChunks({
+          fssUploadId: registration.uploadId,
+          source: source,
+          chunkSize: registration.chunkSize,
+          user: upload.user,
+          onProgress: (bytesUploaded) => onProgress({bytesUploaded, totalBytes: fileSize})
+        });
+      }
     } catch (error) {
       // Ignore cancellation errors
       if (!(error instanceof CancellationError)) {
@@ -230,7 +240,6 @@ export default class FileManagementSystem {
         `Upload cannot be retried if already successful, actual status is ${upload.status}.`
       );
     }
-
     // Attempt to resume an ongoing upload if possible before scraping this one entirely
     const { fssUploadId } = upload.serviceFields;
     let resumeError: Error | undefined;
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,903 +2,909 @@
 # yarn lockfile v1
 
 
-"7zip-bin@~5.1.1":
-  version "5.1.1"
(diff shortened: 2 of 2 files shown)


src/renderer/services/file-management-system/index.ts
```diff
@@ -137,15 +136,26 @@ export default class FileManagementSystem {
...
-
-      // Wait for upload
-      await this.uploadInChunks({
-        fssUploadId: registration.uploadId,
-        source: source,
-        chunkSize: registration.chunkSize,
-        user: upload.user,
-        onProgress: (bytesUploaded) => onProgress({ bytesUploaded, totalBytes: fileSize })
-      });
+      if (registration.chunkStatuses && registration.chunkStatuses[0]){
```
SeanLeRoy, line 139: Does this need to check that the status (at 0) is successful or any specific status or is it just existing good enough?

src/renderer/services/file-management-system/index.ts
```diff
@@ -137,15 +136,26 @@ export default class FileManagementSystem {
...
+        // update the maongodb fua.upload job with the fss uplaod_id field
+        const onRetryProgress = (
+            uploadId: string,
+            progress: UploadProgressInfo
+        ) => {
+          onProgress(progress)
+        };
+        // create the FUA jss uplaod id and info here
```
SeanLeRoy, line 147:
```suggestion
        // create the FUA JSS upload id and info here
```

src/renderer/services/file-management-system/test/file-management-system.test.ts
```diff
@@ -2,23 +2,15 @@ import * as fs from "fs";
 import * as os from "os";
 import * as path from "path";
 
-import { expect } from "chai";
-import { noop } from "lodash";
-import { createSandbox, SinonStubbedInstance } from "sinon";
+import {expect} from "chai";
+import {noop} from "lodash";
+import {createSandbox, SinonStubbedInstance} from "sinon";
```
SeanLeRoy, line 7: Hmm this is not what the other files are formatted like so lets perhaps revert this unless we are committing to using this formatting

src/renderer/services/file-management-system/test/file-management-system.test.ts
```diff
@@ -164,6 +156,97 @@ describe("FileManagementSystem", () => {
...
+        name: "",
+        size: 4,
+        md5: "",
+      });
+
+      // Act
+      await fms.upload(uploadJob, noop);
+
+      // Assert
+      expect(fss.getStatus.called).to.be.true;
```
SeanLeRoy, line 244: What about these assertions tells us a retry for sure happened? If something, could you add a comment saying what it is for future us. If nothing, you can assert what the functions were called with or assert that something wasn't called that is maybe unique to a non-retry code path.

src/renderer/state/template/test/logics.test.ts
```diff
@@ -428,6 +429,8 @@ describe("Template Logics", () => {
         },
       });
       const result: DropResult = {
+        destination: undefined,
+        combine: undefined,
```
SeanLeRoy, line 433: Thanks for addressing this!

SeanLeRoy (changes requested):
Blocking to make sure that test does assert something specific & for the import formatting (see respective comments)

src/renderer/services/file-management-system/index.ts
kmitcham: Just existing should be good enough; the retry/resume code should handle it.

src/renderer/services/file-management-system/test/file-management-system.test.ts
kmitcham: reverted

src/renderer/services/file-management-system/test/file-management-system.test.ts
kmitcham: commented

SeanLeRoy (approved):

src/renderer/state/template/test/logics.test.ts
hughes036: What is going on here?  Does it relate to the main goal of the pr?

src/renderer/services/file-management-system/test/file-management-system.test.ts
hughes036: As you said in the summary, I think this test has some problems.  These assertions strike me as implementation details which could easily change, which makes the test brittle, and it will be confusing to figure out why it broke by a later us.  

Idea: in `retry`, what if a jss job was update with a new `serviceField`, something like `stage: RETRIED`.  The test could then assert that jss was called with exactly that value.

hughes036 (commented):
The change makes sense, I think the test should be improved or if that is too difficult, im not sure it is worth keeping in this brittle state.

src/renderer/services/file-management-system/test/file-management-system.test.ts
kmitcham: Line 252

hughes036 (approved):
Thanks for making that change.

PR #59, opened 2023-04-28
Feature/swe 859

hughes036:
**Objective:**

Implement client side changes needed to surface & test the FSS2 localNASShortcut feature, which is the next phase of solving the memory managements issues plaguing FUA -> FSS2 and making it unusable for multi-TB scale files.

**Changes**

Implemented code paths for conditions that based on a (currently harcoded, later UI driven) flag, will: 
- Bypass the file reading / md5 computing / chunk sending and finalizing paths.  
- Poll `fssStatus.currentFileSize` for progress.
- Execute specific retry code (also bypassing chunked / md5 functionality).

Additional refactoring:
- No longer relying on the FSS2 JSS job for properties which are available from the FSS2Status.
- Unit test improvements

**Testing**
- Unit tests
- Manual (over X terminal)

**How to review**
Start with `FileManagementSystem`

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "16.x",
     "npm": "8.x"
   },
-  "version": "3.0.1",
+  "version": "3.0.2-snapshot.0",
   "build": {
     "appId": "org.aics.alleninstitute.fileupload",
     "files": [
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "16.x",
     "npm": "8.x"
   },
-  "version": "3.0.2-snapshot.6",
+  "version": "3.0.2-snapshot.7",
   "build": {
     "appId": "org.aics.alleninstitute.fileupload",
     "files": [
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "16.x",
     "npm": "8.x"
   },
-  "version": "3.0.2-snapshot.3",
+  "version": "3.0.2-snapshot.4",
   "build": {
     "appId": "org.aics.alleninstitute.fileupload",
     "files": [
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "16.x",
     "npm": "8.x"
   },
-  "version": "3.0.2-snapshot.2",
+  "version": "3.0.2-snapshot.3",
   "build": {
     "appId": "org.aics.alleninstitute.fileupload",
     "files": [
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "16.x",
     "npm": "8.x"
   },
-  "version": "3.0.2-snapshot.0",
+  "version": "3.0.2-snapshot.1",
   "build": {
     "appId": "org.aics.alleninstitute.fileupload",
     "files": [
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "16.x",
     "npm": "8.x"
   },
-  "version": "3.0.2-snapshot.7",
+  "version": "3.0.2-snapshot.8",
   "build": {
     "appId": "org.aics.alleninstitute.fileupload",
     "files": [
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "16.x",
     "npm": "8.x"
   },
-  "version": "3.0.2-snapshot.4",
+  "version": "3.0.2-snapshot.5",
   "build": {
     "appId": "org.aics.alleninstitute.fileupload",
     "files": [
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "16.x",
     "npm": "8.x"
   },
-  "version": "3.0.2-snapshot.8",
+  "version": "3.0.2-snapshot.9",
   "build": {
     "appId": "org.aics.alleninstitute.fileupload",
     "files": [
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "16.x",
     "npm": "8.x"
   },
-  "version": "3.0.2-snapshot.9",
+  "version": "3.0.2-snapshot.10",
   "build": {
     "appId": "org.aics.alleninstitute.fileupload",
     "files": [
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "16.x",
     "npm": "8.x"
   },
-  "version": "3.0.2-snapshot.5",
+  "version": "3.0.2-snapshot.6",
   "build": {
     "appId": "org.aics.alleninstitute.fileupload",
     "files": [
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "16.x",
     "npm": "8.x"
   },
-  "version": "3.0.2-snapshot.1",
+  "version": "3.0.2-snapshot.2",
   "build": {
     "appId": "org.aics.alleninstitute.fileupload",
     "files": [
--- a/src/renderer/services/file-management-system/index.ts
+++ b/src/renderer/services/file-management-system/index.ts
@@ -6,8 +6,6 @@ import * as uuid from "uuid";
 
 import { extensionToFileTypeMap, FileType } from "../../util";
 import FileStorageService, {
-  FSSUpload,
-  RegisterUploadResponse,
   UploadStatus,
   UploadStatusResponse,
 } from "../file-storage-service";
@@ -59,7 +57,7 @@ export default class FileManagementSystem {
     return uuid.v1().replace(/-/g, "");
   }
 
-  private static sleep(timeoutInMs = 2000){
+  private static sleep(timeoutInMs = 2000) {
     return new Promise(resolve => setTimeout(resolve, timeoutInMs))
   }
 
@@ -90,13 +88,13 @@ export default class FileManagementSystem {
         localNasShortcut: true, //TODO accept from ui control chrishu 4/21/23
         ...serviceFields,
       },
-      
+
     });
   }
 
   private async register(
     upload: UploadJob,
-  ): Promise<[RegisterUploadResponse, string, number]> {
+  ): Promise<[UploadStatusResponse, string, number]> {
     // Grab file details
     const source = upload.serviceFields.files[0]?.file.originalPath;
     const fileName = path.basename(source);
@@ -144,14 +142,13 @@ export default class FileManagementSystem {
     onProgress: (progress: UploadProgressInfo) => void
   ): Promise<void> {
     try {
-      const [registration, source, fileSize] = await this.register(upload);
-      if(upload.serviceFields.localNasShortcut){
-        this.updateProgress(registration.uploadId, fileSize, onProgress);
+      const [fssStatus, source, fileSize] = await this.register(upload);
+      if (upload.serviceFields.localNasShortcut) {
+        this.updateProgress(fssStatus.uploadId, fileSize, onProgress);
       } else {
         await this.uploadInChunks({
-          fssUploadId: registration.uploadId,
-          source: source,
-          chunkSize: registration.chunkSize,
+          fssStatus,
+          source,
           user: upload.user,
           onProgress: (bytesUploaded) => onProgress({ bytesUploaded, totalBytes: fileSize })
         });
@@ -245,10 +242,10 @@ export default class FileManagementSystem {
     // Attempt to resume an ongoing upload if possible before scraping this one entirely
     const { fssUploadId } = fuaUpload.serviceFields;
     let resumeError: Error | undefined;
-    
+
     if (fssUploadId) {
       const fssStatus = await this.fss.getStatus(fssUploadId);
-      if(fssStatus?.status !== UploadStatus.INACTIVE){
+      if (fssStatus?.status !== UploadStatus.INACTIVE) {
         try {
           await this.resume(fuaUpload, fssStatus, onProgress);
           return;
@@ -411,26 +408,32 @@ export default class FileManagementSystem {
     if (lastModifiedInMS !== fileLastModifiedInMs) {
       throw new Error("File has been modified since last upload attempt");
     }
-    switch(fssStatus.status){
+    switch (fssStatus.status) {
       case UploadStatus.WORKING:
         // Update status to reflect the resume going smoothly
         await this.jss.updateJob(fuaUpload.jobId, {
           status: JSSJobStatus.RETRYING,
         });
-        if(localNasShortcut) {
-          await this.resumeLocalNasShortcut(fssStatus.uploadId, fuaUpload, fssStatus, onProgress);
+        if (localNasShortcut) {
+          // For localNasShortcut uploads, the way to reume an in progress upload is to call /register on it again. 
+          const [registration, _, fileSize] = await this.register(fuaUpload);
+          this.updateProgress(registration.uploadId, fileSize, (progress: UploadProgressInfo) => onProgress(registration.uploadId, progress))
         } else {
-          await this.resumeChunked(fuaUpload, fssStatus, onProgress);
+          await this.resumeUploadInChunks(fuaUpload, fssStatus, onProgress);
         }
         break;
       case UploadStatus.RETRY:
-        //TODO
+        if (localNasShortcut) {
+          await this.fss.retryFinalizeLocalNasShortcut(fssStatus.uploadId);
+        } else {
+          await this.retryChunkedUpload(fssStatus);
+        }
         break;
       case UploadStatus.COMPLETE:
         // If an FSS upload status is complete it has performed everything
         // it needs to and may just need the client to finish its portion
         const { fileId } = fssStatus;
-        if(!fileId){
+        if (!fileId) {
           throw new Error("FileId was not published on COMPLETE upload: " + fssStatus.uploadId)
         }
         await this.complete(fuaUpload, fileId)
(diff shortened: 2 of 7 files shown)


src/renderer/services/file-management-system/index.ts
```diff
@@ -485,15 +533,16 @@ export default class FileManagementSystem {
...
-    chunkSize: number,
     user: string,
     onProgress: (bytesUploaded: number) => void,
     initialChunkNumber?: number,
     partiallyCalculatedMd5?: string,
   }): Promise<void> {
-    const { fssUploadId, source, chunkSize, user, onProgress, initialChunkNumber = 0, partiallyCalculatedMd5 } = config;
+    const { fssStatus, source, user, onProgress, initialChunkNumber = 0, partiallyCalculatedMd5 } = config;
+    const fssUploadId = fssStatus.uploadId;
+    const chunkSize = fssStatus.chunkSize;
```
SeanLeRoy, line 545: Is this chunkSize guaranteed to be the chunkSize at the time of the original upload's start? Like if we upgrade FSS after an upload has been attempted could the chunkSize change? That is the reasoning behind why it previously relied on the serviceFields. If we can just rely on FSS, lets remove it from the serviceField if possible.

src/renderer/services/file-management-system/index.ts
```diff
@@ -547,13 +596,14 @@ export default class FileManagementSystem {
       chunkNumber += 1;
       uploadChunkPromises.push(uploadChunk(chunk, chunkNumber, md5ThusFar));
     }
-
+    const offset = chunkSize * initialChunkNumber;
+    //TODO SWE-865 only read if offset < fssStatus
```
SeanLeRoy, line 600: Would this codepath be triggered under this condition?

src/renderer/services/file-management-system/index.ts
```diff
@@ -398,85 +397,134 @@ export default class FileManagementSystem {
...
+  }
+
+  private async updateProgress(
+    fssUploadId: string,
+    fileSize: number,
+    onProgress: (progress: UploadProgressInfo) => void) {
+    const fssStatusResponse = await this.fss.getStatus(fssUploadId);
+    let fssStatus = fssStatusResponse?.status;
+    while (fssStatus !== UploadStatus.COMPLETE) {
+      await FileManagementSystem.sleep(1000); //TODO too short?  Tests timeout if > 2 sec
```
SeanLeRoy, line 512: I personally prefer tracking TODOs like this that are easily resolvable outside of the codebase or at least outside the PR.

src/renderer/services/file-management-system/index.ts
```diff
@@ -398,85 +397,134 @@ export default class FileManagementSystem {
...
+      fssStatus = fssStatusResponse?.status;
+      switch (fssStatus) {
+        case UploadStatus.WORKING:
+          onProgress({ bytesUploaded: fssStatusResponse.currentFileSize, totalBytes: fileSize });
+          break;
+        case UploadStatus.POST_PROCESSING:
+          onProgress({ bytesUploaded: fssStatusResponse.currentFileSize, totalBytes: fileSize });
+          break;
+        case UploadStatus.INACTIVE:
+        case UploadStatus.RETRY:
```
SeanLeRoy, line 523: Based on the logic in the resume function I believe we can assume the finalize step needs to be retried (I have another comment double checking that). Since that is actionable by the user, might be worth changing the error statement a little for the retry case.

src/renderer/services/file-management-system/index.ts
```diff
@@ -398,85 +397,134 @@ export default class FileManagementSystem {
...
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          const [registration, _s, fileSize] = await this.register(fuaUpload);
+          this.updateProgress(registration.uploadId, fileSize, (progress: UploadProgressInfo) => onProgress(registration.uploadId, progress))
+        } else {
+          await this.resumeUploadInChunks(fuaUpload, fssStatus, onProgress);
+        }
+        break;
+      case UploadStatus.RETRY:
+        if (localNasShortcut) {
+          await this.fss.retryFinalizeLocalNasShortcut(fssStatus.uploadId);
```
SeanLeRoy, line 427: I understand that if a chunk needs to be retried the whole job will present as `WORKING` still in the chunked scenario, but what happens in the local NAS shortcut scenario when something goes awry during the copy? Does it just get marked as `INACTIVE`?

SeanLeRoy (changes requested):
Looks great! Requesting change since I just have a lot of questions, likely fine though!

src/renderer/services/file-management-system/index.ts
```diff
@@ -86,11 +85,49 @@ export default class FileManagementSystem {
       serviceFields: {
         files: [metadata],
         type: "upload",
+        localNasShortcut: true, //TODO accept from ui control chrishu 4/21/23
```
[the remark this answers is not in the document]
hughes036: [SWE-866](https://aicsjira.corp.alleninstitute.org/browse/SWE-866)

src/renderer/services/file-management-system/index.ts
```diff
@@ -222,60 +229,52 @@ export default class FileManagementSystem {
...
-        // No-op: move on if this failed
-      }
 
-      if (
-        fssStatus?.status === UploadStatus.WORKING ||
-        fssStatus?.status === UploadStatus.RETRY ||
-        fssStatus?.status === UploadStatus.COMPLETE
-      ) {
+    if (fssUploadId) {
+      const fssStatus = await this.fss.getStatus(fssUploadId);
```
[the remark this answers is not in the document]
hughes036: What is the value in the no-op?

Since there is no dependency on it, I see having the `fssStatus` variable at all as dead code.

src/renderer/services/file-management-system/index.ts
hughes036: `RETRY` is for when a post-upload task fails (and exceeds retries, like if MMS is down).  The localNasShortcut 'copy' is more of a read (by FSS2), and then upload chunks using the existing infrastructure.  If something goes awry, there will just be missing chunks (just like with the chunked upload).

src/renderer/services/file-management-system/index.ts
```diff
@@ -398,85 +397,134 @@ export default class FileManagementSystem {
...
-        await this.jss.updateJob(upload.jobId, {
-          status: JSSJobStatus.RETRYING,
-        });
+        await this.complete(fuaUpload, fileId)
+        break;
+      case UploadStatus.POST_PROCESSING:
+        onProgress(fssStatus.uploadId, { bytesUploaded: fssStatus.currentFileSize, totalBytes: fssStatus.currentFileSize });
+        break;
+      default:
+        console.error(`Unexpected FSS UploadStatus encountered: ${fssStatus?.status}`);
```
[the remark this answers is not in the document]
hughes036: I agree about throwing an error, but if this is reached, that means they are using a client version that is out of date with the FSS2 API.  They should be prompted to update to the latest version, right?

src/renderer/services/file-management-system/index.ts
```diff
@@ -398,85 +397,134 @@ export default class FileManagementSystem {
...
-            source: originalPath,
-            chunkSize: fssStatus.chunkSize,
-            user: upload.user,
-            onProgress: (bytesUploaded) =>
-              onProgress(upload.jobId, { bytesUploaded, totalBytes: fileSize }),
-            initialChunkNumber: lastChunkNumber,
-            partiallyCalculatedMd5
-          });
-        }
+  private async retryChunkedUpload(
```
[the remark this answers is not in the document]
hughes036: renamed

src/renderer/services/file-management-system/index.ts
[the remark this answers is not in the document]
hughes036: I had to make it short in order for the tests to pass, but it must have been in a bad state then because I just upped to 5 sec and they are now passing.

hughes036: I like TODOs because stopping to create a JIRA ticket breaks me out of the flow.  I go back and create the tickets later when i make a pr.  But I like to keep to TODOs because they are a placeholder and create a sense of pressure to go back and actually do it, rather than let it be forgotten in the backlog.

package.json
```diff
@@ -4,7 +4,7 @@
     "node": "16.x",
     "npm": "8.x"
   },
-  "version": "3.0.1",
+  "version": "3.0.2-snapshot.3",
```
griffinfujioka, line 7: I don't know this codebase well, but double checking to ensure the `snapshot` here is intended.

griffinfujioka (approved):

src/renderer/services/file-management-system/index.ts
hughes036: The behavior of the tests suggests it would.  I tried asserting that `fileReader.read` is not called and that failed.

src/renderer/services/file-management-system/index.ts
hughes036: I can't think of anything too terrible happening.  The only reasons I know or that chunkSize matters are:
- performance
- keeping the total number of chunks below the S3 limit.

If we make a perf related change on the backend, I dont see an upload resuming at the new size being a problem.  If it is related to the S3 limit, then something must have changed with S3 usage and continuing under the old assumption seems unwise.  What do you think?

package.json
hughes036: It gets editted by `yarn version --prerelease --preid=snapshot`.  Perhaps this shouldnt be versioned with a snapshot build though.  @SeanLeRoy ?

src/renderer/services/file-management-system/index.ts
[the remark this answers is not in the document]
SeanLeRoy: >They should be prompted to update to the latest version, right?

Yes though not because of this error, if they have gotten here they have already ignored that prompt. Regardless, I think throwing an error here is more visible

src/renderer/services/file-management-system/index.ts
SeanLeRoy: It would be a problem due to how we calculate the offset of where we left off in the upload. If we uploaded 5 500mb chunks (2500mb) then change the chunksize to 800 on the backend we would resume at 4000mb

package.json
SeanLeRoy: Its fine since it reflects reality

src/renderer/services/file-management-system/index.ts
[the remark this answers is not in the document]
SeanLeRoy: Sweet. Did you get a chance to test the many uploads at once?

src/renderer/services/file-management-system/index.ts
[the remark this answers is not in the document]
hughes036: > > Since there is no dependency on it, I see having the fssStatus variable at all as dead code.
> 
> It is used right below this...?

Ok, i was thinking about something else when I wrote that.  Nm.



> > What is the value in the no-op?
> 
> So that if/when we resume we could enter the try/catch and have the upload get canceled if we can't resume. After thinking about this more, I'd just expand the try/catch to include this request.

I'll look at this.

src/renderer/services/file-management-system/index.ts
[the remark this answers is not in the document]
hughes036: Nope.  I'll try it.

src/renderer/services/file-management-system/index.ts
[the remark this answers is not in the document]
hughes036: Converted to an error.

About the version prompt, i was responded to your idea that they should be prompted to retry.  If their version is out of date, then retrying again is probably not advisable until they update the version.

src/renderer/services/file-management-system/index.ts
hughes036: Actually, the chunksize being passed in here is from the the FSS status, not the service field.

SeanLeRoy: >Actually, on main the chunksize being passed in here is also from the the FSS status, not the service field.

Ah so its been a potential problem for a while, could be too niche to matter then

pgarrison (approved):
All looks good to me. We're probably going to need some future work to make the retry/resume logic more robust (both in FUA and FSS), but that can be a later ticket

src/renderer/services/file-management-system/index.ts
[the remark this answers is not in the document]
hughes036: Discussed in personf, implented change

src/renderer/services/file-management-system/index.ts
hughes036: backlog ticket

SeanLeRoy (approved):

kmitcham (approved):

src/renderer/services/file-management-system/index.ts
```diff
@@ -86,11 +85,49 @@ export default class FileManagementSystem {
...
+    const { size: fileSize, mtime: fileLastModified } = await fs.promises.stat(source);
+    const fileLastModifiedInMs = fileLastModified.getTime();
+    // Heuristic which in most cases, prevents attempting to upload a duplicate
+    if (await this.fss.fileExistsByNameAndSize(fileName, fileSize)) {
+      throw new Error(
+        `File ${fileName} with size ${fileSize} already exists in FMS`
+      );
+    }
+
+    // Start job in FSS
```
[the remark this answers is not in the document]
hughes036: Look like no.  removed.

src/renderer/services/file-management-system/index.ts
hughes036: [SWE-879](https://aicsjira.corp.alleninstitute.org/browse/SWE-879)

src/renderer/services/file-management-system/index.ts
```diff
@@ -78,6 +77,7 @@ export default class FileManagementSystem {
     user: string,
     serviceFields: Partial<UploadServiceFields> = {}
   ): Promise<UploadJob> {
+    console.log(metadata.file.originalPath); // TODO remove after pilot testing
```
SeanLeRoy, line 80: Is this TODO and similar for returning to after merge?

src/renderer/services/file-management-system/index.ts
```diff
@@ -86,11 +86,50 @@ export default class FileManagementSystem {
       serviceFields: {
         files: [metadata],
         type: "upload",
+        // localNasShortcut: true, // TODO for reliable pilot testing
```
SeanLeRoy: this line seems deletable

src/renderer/services/file-management-system/index.ts
```diff
@@ -101,51 +140,22 @@ export default class FileManagementSystem {
...
-
-      // Wait for upload
-      await this.uploadInChunks({
-        fssUploadId: registration.uploadId,
-        source: source,
-        chunkSize: registration.chunkSize,
-        user: upload.user,
-        onProgress: (bytesUploaded) => onProgress({ bytesUploaded, totalBytes: fileSize })
-      });
+      //poll->onProgress
```
SeanLeRoy, line 157: ?

src/renderer/services/file-management-system/index.ts
```diff
@@ -398,85 +407,137 @@ export default class FileManagementSystem {
...
+    switch (fssStatus.status) {
+      case UploadStatus.WORKING:
+        // Update status to reflect the resume going smoothly
+        await this.jss.updateJob(fuaUpload.jobId, {
+          status: JSSJobStatus.RETRYING,
+        });
+        if (localNasShortcut) {
+          // For localNasShortcut uploads, the way to reume an in progress upload is to call /register on it again. 
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          const [registration, _s, fileSize] = await this.register(fuaUpload);
```
SeanLeRoy, line 428: Hmm if you make it just `_` the eslint error might go away automatically

src/renderer/services/file-management-system/test/file-management-system.test.ts
```diff
@@ -82,6 +82,74 @@ describe("FileManagementSystem", () => {
...
+
+      // Act
+      await fms.upload(upload, noop);
+
+      // Assert
+      expect(fileReader.read).to.have.been.calledOnce;
+      expect(fss.finalize.calledOnceWithExactly(uploadId, expectedMd5)).to.be.true;
+    });
+
+    it("Inits local_nas_shortcut instead of chunked.", async () => {
```
SeanLeRoy, line 117: I think a test where fssStatus is returned as `INACTIVE` and/or `RETRY` to exercise an error case would be useful

SeanLeRoy (approved):

src/renderer/services/file-management-system/index.ts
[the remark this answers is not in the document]
hughes036: I did my best under the conditions Ive been relegated to (testing over x term with crazy lag, testing very small files).  Im gonna plan on organizing an on prem testing session with help from someone (hopefully @kmitcham  or @pgarrison )

src/renderer/services/file-management-system/index.ts
hughes036: Before - after testing a snapshot version on windows.  just wanna make sure the path look the way i expect, and implement some unit tests.

src/renderer/services/file-management-system/index.ts
hughes036: im uncommenting that for pilot testing on windows.

src/renderer/services/file-management-system/index.ts
hughes036: removed.

PR #60, opened 2023-05-11
Bugfix/swe 826 2

hughes036:
TODO: @kmitcham have a look at this, and lmk if we need the change in the logics test from (I dont understand the context):

https://github.com/aics-int/aics-file-upload-app/pull/57/files#diff-21ed2ce30a3cde6a026c8db1c356ef67558715bdc373bfa0781af33a78ad1969

### DIFF
--- a/src/renderer/services/file-management-system/index.ts
+++ b/src/renderer/services/file-management-system/index.ts
@@ -6,8 +6,6 @@ import * as uuid from "uuid";
 
 import { extensionToFileTypeMap, FileType } from "../../util";
 import FileStorageService, {
-  FSSUpload,
-  RegisterUploadResponse,
   UploadStatus,
   UploadStatusResponse,
 } from "../file-storage-service";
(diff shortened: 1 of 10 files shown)


src/renderer/services/file-management-system/index.ts
```diff
@@ -101,6 +101,8 @@ export default class FileManagementSystem {
     const mntPointForcedLowerCase = source.replace(/allen/gi, 'allen'); // Windows is inconsistent here (have seen both 'ALLEN' and Allen' generated in the wild), 
                                                                         // and unix paths are case sensitive.
     const normalizedPath = path.normalize(mntPointForcedLowerCase);     // Drop proceeding / from //allen.
+    console.log("Source path: " + source);
```
kmitcham, line 104: TODO: remove these console logs?

kmitcham (approved):
Looks like we've covered the case here.

src/renderer/services/file-management-system/index.ts
hughes036: Yep, did so upon rebasing onto my other branch.

PR #61, opened 2023-05-31
status/progress updates from JSS sse

hughes036:
**Summary**
This pr is focused on converting upload progress updated from being based on callback and polling, to being based on Server Sent Events.  The result is simpler and easier to read.

**Related:**
https://github.com/aics-int/file-storage-service/pull/233

**How to review**
Start with `renderer/containers/App/index.tsx`, then `renderer/state/job/logics.ts` and `src/renderer/state/upload/logics.ts`.  Then, `src/renderer/services/file-management-system/index.ts.`

**Testing** 
* unit testing
* manual testing

### DIFF
--- a/src/renderer/containers/App/index.tsx
+++ b/src/renderer/containers/App/index.tsx
@@ -64,11 +64,11 @@ message.config({
 });
 
 export function handleUploadJobUpdates(job: JSSJob, dispatch: Dispatch<any>){
-      // An FSS job update happens when:
-      //   * fileId has been published
-      //   * progress has been published on a pre-upload md5, file upload, or post-upload md5
-      //   * requires this clients intervention to retry
       if (job.service === Service.FILE_STORAGE_SERVICE) {
+        // An FSS job update happens when:
+        //   * fileId has been published
+        //   * progress has been published on a pre-upload md5, file upload, or post-upload md5
+        //   * requires this clients intervention to retry
         const fssJob = job as FSSUpload;
         const totalBytes = fssJob.serviceFields.fileSize || 0; // 0 is a safe default, but in practice filesize is initialized immediately after job creation.
         if (fssJob.serviceFields?.fileId || fssJob.currentStage === UploadStatus.INACTIVE || fssJob.currentStage === UploadStatus.RETRY) {
--- a/src/renderer/containers/App/index.tsx
+++ b/src/renderer/containers/App/index.tsx
@@ -117,19 +117,19 @@ export default function App() {
       // the addition of the fileId i.e. FSS's completion, has failed,
       // or requires this clients intervention to retry
       if (
-        job.service === Service.FILE_STORAGE_SERVICE &&
-        (job.serviceFields?.fileId || job.status === JSSJobStatus.FAILED || job.currentStage === UploadStatus.RETRY)
-      ) {
-        dispatch(receiveFSSJobCompletionUpdate(job as FSSUpload));
-      } else if (job.serviceFields?.type === "upload") {
+        job.service === Service.FILE_STORAGE_SERVICE) {
+          if (job.serviceFields?.fileId || job.status === JSSJobStatus.FAILED || job.currentStage === UploadStatus.RETRY) {
+            dispatch(receiveFSSJobCompletionUpdate(job as FSSUpload));
+          } else if(job.serviceFields?.preUploadMd5 && (job.serviceFields.preUploadMd5 != job.serviceFields?.fileSize)) {
+            dispatch(updateUploadProgressInfo(job.jobId, { bytesUploaded: job.serviceFields?.preUploadMd5, totalBytes: job.serviceFields?.fileSize, step: Step.ONE }));
+          } else if(job.serviceFields?.currentFileSize && (job.serviceFields.currentFileSize!= job.serviceFields?.fileSize)) {
+            dispatch(updateUploadProgressInfo(job.jobId, { bytesUploaded: job.serviceFields?.currentFileSize, totalBytes: job.serviceFields?.fileSize, step: Step.TWO }));
+          } else if(job.serviceFields?.postUploadMd5 && (job.serviceFields.postUploadMd5 != job.serviceFields?.fileSize)) {
+            dispatch(updateUploadProgressInfo(job.jobId, { bytesUploaded: job.serviceFields?.postUploadMd5, totalBytes: job.serviceFields?.fileSize, step: Step.THREE }));
+          }
+        } else if (job.serviceFields?.type === "upload") {
         // Otherwise separate user's other jobs from ones created by this app
         dispatch(receiveJobUpdate(job as UploadJob));
-      } else if(job.serviceFields?.pre_upload_md5 != job.serviceFields?.file_size) {
-        dispatch(updateUploadProgressInfo(job.jobId, { bytesUploaded: job.serviceFields?.pre_upload_md5, totalBytes: job.serviceFields?.file_size, step: Step.ONE }));
-      } else if(job.serviceFields?.current_file_size != job.serviceFields?.file_size) {
-        dispatch(updateUploadProgressInfo(job.jobId, { bytesUploaded: job.serviceFields?.pre_upload_md5, totalBytes: job.serviceFields?.file_size, step: Step.TWO }));
-      } else if(job.serviceFields?.post_upload_md5 != job.serviceFields?.file_size) {
-        dispatch(updateUploadProgressInfo(job.jobId, { bytesUploaded: job.serviceFields?.pre_upload_md5, totalBytes: job.serviceFields?.file_size, step: Step.THREE }));
       }
     });
 
--- a/src/renderer/containers/App/index.tsx
+++ b/src/renderer/containers/App/index.tsx
@@ -64,9 +64,10 @@ message.config({
 });
 
 export function handleUploadJobUpdates(job: JSSJob, dispatch: Dispatch<any>){
-        // An FSS job update is only important to us when it is signaling
-      // the addition of the fileId i.e. FSS's completion, has failed,
-      // or requires this clients intervention to retry
+      // An FSS job update happens when:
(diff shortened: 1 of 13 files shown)


griffinfujioka (approved):

src/renderer/containers/App/index.tsx
```diff
@@ -115,11 +117,17 @@ export default function App() {
...
       if (
-        job.service === Service.FILE_STORAGE_SERVICE &&
-        (job.serviceFields?.fileId || job.status === JSSJobStatus.FAILED || job.currentStage === UploadStatus.RETRY)
-      ) {
-        dispatch(receiveFSSJobCompletionUpdate(job as FSSUpload));
-      } else if (job.serviceFields?.type === "upload") {
+        job.service === Service.FILE_STORAGE_SERVICE) {
+          if (job.serviceFields?.fileId || job.status === JSSJobStatus.FAILED || job.currentStage === UploadStatus.RETRY) {
+            dispatch(receiveFSSJobCompletionUpdate(job as FSSUpload));
+          } else if(job.serviceFields?.preUploadMd5 && (job.serviceFields.preUploadMd5 !== job.serviceFields?.fileSize)) {
```
SeanLeRoy, line 123: Are `job.serviceFields.preUploadMd5` and `job.serviceFields.postUploadMd5` file sizes not MD5 hashes?

src/renderer/containers/App/index.tsx
```diff
@@ -115,11 +117,17 @@ export default function App() {
...
+            dispatch(updateUploadProgressInfo(job.jobId, { bytesUploaded: job.serviceFields?.preUploadMd5, totalBytes: job.serviceFields?.fileSize, step: Step.ONE }));
+          } else if(job.serviceFields?.currentFileSize && (job.serviceFields.currentFileSize!== job.serviceFields?.fileSize)) {
+            dispatch(updateUploadProgressInfo(job.jobId, { bytesUploaded: job.serviceFields?.currentFileSize, totalBytes: job.serviceFields?.fileSize, step: Step.TWO }));
```
SeanLeRoy, line 126: This works with the non-shortcut upload too right?

src/renderer/services/file-management-system/index.ts
```diff
@@ -145,6 +147,16 @@ export default class FileManagementSystem {
         lastModifiedInMS: fileLastModifiedInMs,
       },
     });
+    switch (registration.status) {
+      case UploadStatus.RETRY:
+        throw new Error(
+          'Your file uploaded and is safe, but a post-upload task failed on the backend.  Please wait a few moments, and the retry (select the box to the right of your upload, and use the "retry" button.'
```
SeanLeRoy, line 153: Is it uploaded after just registering with FSS?

src/renderer/services/file-management-system/index.ts
```diff
@@ -157,29 +169,18 @@ export default class FileManagementSystem {
...
-        ) => {
-          onProgressInfo(progress)
-        };
-        await this.retry(upload.jobId, onRetryProgress);                  // create the FUA JSS upload id and info here                  
-      } else {
+      const [fssStatus, source] = await this.register(upload);
+      if(fssStatus.chunkStatuses && fssStatus.chunkStatuses[0]) {  
+        //Handles the case where FUA believes this is a new upload, 
+        //but actually, it is partially complete already.
+        await this.retry(upload.jobId);                  // create the FUA JSS upload id and info here
```
SeanLeRoy, line 178: I'm confused by this comment, the FUA JSS upload ID has to exist before this method is called.

src/renderer/services/file-management-system/index.ts
```diff
@@ -157,29 +169,18 @@ export default class FileManagementSystem {
...
-        const onRetryProgress = (                                         // Update the JSS FUA upload job with the FSS upload ID field
-          uploadId: string,
-          progress: UploadProgressInfo
-        ) => {
-          onProgressInfo(progress)
-        };
-        await this.retry(upload.jobId, onRetryProgress);                  // create the FUA JSS upload id and info here                  
-      } else {
+      const [fssStatus, source] = await this.register(upload);
+      if(fssStatus.chunkStatuses && fssStatus.chunkStatuses[0]) {
```
SeanLeRoy, line 175: Can you comment on the cases when this if would be true? I see line 176, but does that mean we would enter this on every local NAS upload?

src/renderer/services/file-storage-service/index.ts
```diff
@@ -11,6 +11,10 @@ import { HttpClient } from "../types";
 export interface FSSUpload extends JSSJob {
   serviceFields: {
     fileId?: string;
+    pre_upload_md5?: string;
```
SeanLeRoy, line 14: Sort of following up on another comment I made in App/index.tsx, if this is a string would it ever equal an integer like file size?

src/renderer/containers/App/index.tsx
hughes036: The are the number of bytes that have been read so far while computing the md5 hashes (the progress).

src/renderer/containers/App/index.tsx
hughes036: right

src/renderer/services/file-storage-service/index.ts
hughes036: hm, that should be a number.

src/renderer/services/file-management-system/index.ts
hughes036: localNas uploads also create chunks (the file is just read server side).  So If the server restarted 1/2 through either type of upload, and then FUA retried, this block would execute.

src/renderer/services/file-management-system/index.ts
hughes036: Removed.
I think that was a merge artifact.  this code came from a parallel branch that @kmitcham worked on.

src/renderer/services/file-management-system/index.ts
hughes036: For a localNas upload, yes.  
But good catch, i need to think about this a bit.

hughes036: This came from the `waitForServerCopy` method which previously was called just after registering.  It checked for these statuses since the upload could transition into them while FUA polled for the status.  I now realize is doesnt make sense to check for it right after registering.  The problem now is that if the upload does transition into these states, fua will simply stop receiving progress updates.  It won't know that it happened, since it is not polling.  I guess the solution is to add a serviceField for fssStatus, so that FUA recieves a SSE when this happens.

hughes036: Oh right! it does update JSS, in the stage field.  So I just need to dispatch an event to fail the FUA upload in those cases so that the user can retry later.

hughes036: Looks like this is possibly handled already, but i need some in person review.

src/renderer/containers/App/index.tsx
hughes036: Btw, do you have ideas on how to unit this this code?  verify that the actions are dispatched?  I havent seen an example in this codebase.

kmitcham (approved):
sorry for delay, lost this tab for a while

src/renderer/services/file-management-system/index.ts
hughes036: discussed in person, refactored

src/renderer/containers/App/index.tsx
```diff
@@ -62,6 +63,29 @@ message.config({
   maxCount: 1,
 });
 
+export function handleUploadJobUpdates(job: JSSJob, dispatch: Dispatch<any>){
+      // An FSS job update happens when:
```
SeanLeRoy, line 67: This is a bit pedantic so probably ignore, but I'd imagine the job changes in other scenarios too right? We just only care about the 3 you described.

src/renderer/containers/Table/CustomCells/StatusCell/test/StatusCell.test.tsx
```diff
@@ -150,8 +155,8 @@ describe("<StatusCell />", () => {
 
     // Assert
     expect(wrapper.find(Tooltip).prop("title")).to.equal(
-      "WORKING - Step 2 of 2: Adding metadata"
+      "WORKING - Step 3 of 3: Post-upload tasks"
     );
-    expect(wrapper.find(Progress).prop("percent")).to.equal(0);
+    // expect(wrapper.find(Progress).prop("percent")).to.equal(0);
```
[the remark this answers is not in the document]
hughes036: a mistake, good catch

src/renderer/state/job/logics.ts
```diff
@@ -168,30 +178,10 @@ const receiveFSSJobCompletionUpdateLogics = createLogic({
           fssUpload.serviceFields?.fileId as string
         );
       } else if (
-        fssUpload.status === JSSJobStatus.FAILED &&
-        matchingUploadJob.status !== JSSJobStatus.FAILED
+        jobShouldFailAccordingToFSSJobStage &&
```
SeanLeRoy, line 181: Readability 🚀

SeanLeRoy (approved):
LGTM besides that commented out thingy Kevin mentioned.

src/renderer/containers/Table/CustomCells/StatusCell/test/StatusCell.test.tsx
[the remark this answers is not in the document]
hughes036: deleted; % complete is no longer bound to Step.

src/renderer/containers/App/index.tsx
hughes036: How about if i move the comment to inside the condition?
`      if (job.service === Service.FILE_STORAGE_SERVICE) {
`

hughes036: pushed

Issue #68, opened 2024-03-19
Support multifile uploads


tyler-foster:
As a user, I want to be able to upload "multifiles", which are files that are effectively directories with specific structures, e.g. `.zarr` and `.sldy` files. These uploads should present as a single entry in the File Upload App's "Uploads" table and should have a single FMS FileId associated with them.

PR #69, opened 2024-03-27
[#68] Add initial structure for supporting .sldy, .zarr uploads

tyler-foster:
### Context
For a long time now, the science teams have wanted to upload particular file types to FMS - those file types are (at minimum) `.sldy` and `.zarr` files, which are "multifiles", AKA special directory structures that may be interpreted as a single file format.

We've gotten to the point where FSS can support multifile uploads, so naturally the next step is to update our FSS clients to do the same.

### Changes
* Added an expectation of an envvar called `USE_MULTIFILE_ASSUMPTION` to toggle multifile support on and off
   *  For testing purposes until multifile support is 100% ready for users to get their hands on, per request.
* When multifile support is enabled, directory uploads are checked to see if the name of the given directory ends in an expected file extension. Multifiles then appear as a single entry in the "My Uploads" table.
   * I have not interviewed people to find out if there are any multifile formats for us to expect beyond `.zarr` and `.sldy`, but adding to the list should be easy.
 *  If an upload is determined to be a multifile, then the FSS `/register` requests have `multifile: true` added to them
 * Multifile uploads have `multifile: true` in their JSS job `serviceFields`, which is then read in `fms.register()`. This should mean upload retries will work correctly.

### Testing
* Added unit tests for new `util` functions
* Manual testing via the UI
  * If you want to do this yourself, flip the boolean mentioned above -> run the app locally (`yarn dev`) -> upload any directory named `<whatever>.sldy` or `.zarr`.

### Reviewing
The changes really aren't too big. For review, I would probably check out the new methods in `src/renderer/util/index.ts` first, then look at `state/upload/logics.ts`, then the FMS / FSS service files. 

### TODO
* In this PR
  * There may be a straggling TODO I left in here.
  * Comment for one of the functions I touched in `util`.
  * Any unit test ideas that come to mind.
* As separate work
  * Multifile progress reporting in FSS and the upload app 
  * Talking to users to make sure multifile uploads work as they would expect.

### Notes
There's something smelly about the upload logic structure in the FUA - I haven't quite mapped it out yet, but I think we could re-use more of our code between uploading files with or without annotations / metadata.

### DIFF
--- a/src/renderer/services/file-storage-service/index.ts
+++ b/src/renderer/services/file-storage-service/index.ts
@@ -113,7 +113,7 @@ export default class FileStorageService extends HttpCacheClient {
       file_size: fileSize,
       local_nas_shortcut: localNasPath !== undefined,
       local_nas_path: localNasPath,
-      multifile :!!isMultifile,
+      multifile: !!isMultifile,
     };
     return this.post<UploadStatusResponse>(
       url,
--- a/src/renderer/services/file-storage-service/index.ts
+++ b/src/renderer/services/file-storage-service/index.ts
@@ -113,7 +113,7 @@ export default class FileStorageService extends HttpCacheClient {
       file_size: fileSize,
       local_nas_shortcut: localNasPath !== undefined,
       local_nas_path: localNasPath,
-      multifile: isMultifile !== undefined ? isMultifile : false,
+      multifile :!!isMultifile,
     };
     return this.post<UploadStatusResponse>(
       url,
--- a/src/renderer/services/file-storage-service/index.ts
+++ b/src/renderer/services/file-storage-service/index.ts
@@ -100,6 +100,7 @@ export default class FileStorageService extends HttpCacheClient {
     fileType: FileType,
     fileSize: number,
     localNasPath?: string,
+    isMultifile?: boolean,
   ): Promise<UploadStatusResponse> {
     const url = `${FileStorageService.BASE_UPLOAD_PATH}/register`;
     const postBody = {
@@ -111,7 +112,8 @@ export default class FileStorageService extends HttpCacheClient {
       // so the conversion must be manual each request
       file_size: fileSize,
       local_nas_shortcut: localNasPath !== undefined,
-      local_nas_path: localNasPath
+      local_nas_path: localNasPath,
+      multifile: isMultifile !== undefined ? isMultifile : false,
     };
     return this.post<UploadStatusResponse>(
       url,
--- a/src/renderer/state/upload/logics.ts
+++ b/src/renderer/state/upload/logics.ts
@@ -726,11 +726,9 @@ const uploadWithoutMetadataLogic = createLogic({
 
     let uploads: UploadJob[];
     try {
-      const filePaths = await Promise.all(deps.action.payload.map(async (filePath) => {
-        const isMultifile = determineIsMultifile(filePath);
-        // todo determineFilesFromNestedPaths first arg may not need to be an array?
-        return await determineFilesFromNestedPaths([filePath], isMultifile);
-      }));
+      const filePaths = await determineFilesFromNestedPaths(
+        deps.action.payload
+      );
       uploads = await Promise.all(
           filePaths.flat().map((filePath) => {
             const isMultifile = determineIsMultifile(filePath); // todo: simplify this whole function so this isn't called twice
--- a/src/renderer/state/upload/logics.ts
+++ b/src/renderer/state/upload/logics.ts
@@ -12,7 +12,13 @@ import FileManagementSystem from "../../services/file-management-system";
 import {UploadJob} from "../../services/job-status-service/types";
 import {AnnotationType, ColumnType} from "../../services/labkey-client/types";
 import {Template} from "../../services/metadata-management-service/types";
-import {determineFilesFromNestedPaths, extensionToFileTypeMap, FileType, splitTrimAndFilter} from "../../util";
+import {
+  determineFilesFromNestedPaths,
+  determineIsMultifile,
+  extensionToFileTypeMap,
+  FileType,
+  splitTrimAndFilter
+} from "../../util";
 import {requestFailed} from "../actions";
 import {setErrorAlert} from "../feedback/actions";
 import {setPlateBarcodeToPlates} from "../metadata/actions";
@@ -687,23 +693,6 @@ const submitFileMetadataUpdateLogic = createLogic({
   },
   type: SUBMIT_FILE_METADATA_UPDATE,
 });
-
-/**
- * Use a given filepath's extension to determine if it is a "multifile".
- * @param filePath Path to the file
- */
-function determineIsMultifile(filePath: string): boolean {
-  const multifileExtensions = ['.zarr', '.sldy'];
-  const combinedExtensions = multifileExtensions.join('|');
-
-  // "ends with one of the listed extensions, ignoring casing"
-  // otherwise written like: /.zarr|.sldy$/i
-  const matcher = new RegExp(combinedExtensions + "$", 'i');
-
-  // If the regex matches it will return an array (truthy).
-  // If the regex doesn't match it will return null (falsy).
-  return Boolean(filePath.match(matcher));
-}
 
 const uploadWithoutMetadataLogic = createLogic({
   process: async (
--- a/src/renderer/state/upload/logics.ts
+++ b/src/renderer/state/upload/logics.ts
@@ -1,29 +1,21 @@
 import * as fs from "fs";
 import * as path from "path";
 
-import {
-  castArray,
-  flatMap,
-  forEach,
-  get,
-  isNil,
-  trim,
-} from "lodash";
-import { isDate, isMoment } from "moment";
-import { createLogic } from "redux-logic";
+import {castArray, flatMap, forEach, get, isNil, trim,} from "lodash";
+import {isDate, isMoment} from "moment";
+import {createLogic} from "redux-logic";
 
-import { RendererProcessEvents } from "../../../shared/constants";
-import { AnnotationName, LIST_DELIMITER_SPLIT } from "../../constants";
+import {RendererProcessEvents} from "../../../shared/constants";
+import {AnnotationName, LIST_DELIMITER_SPLIT} from "../../constants";
 import BatchedTaskQueue from "../../entities/BatchedTaskQueue";
-import FileManagementSystem, {
-} from "../../services/file-management-system";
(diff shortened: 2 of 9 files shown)


src/renderer/util/index.ts
```diff
@@ -1,11 +1,21 @@
...
 import { flatten, memoize, uniq } from "lodash";
 
 import { LIST_DELIMITER_SPLIT, MAIN_FONT_WIDTH } from "../constants";
 
+///////////////////////////////////////////////////////////////////
+// TODO - Remove this once multifile support is feature-complete //
+//         Flip the boolean to "true" for testing                //
+///////////////////////////////////////////////////////////////////
+
+const USE_MULTIFILE_ASSUMPTION = false;
```
kmitcham, line 15: Hmm.  could this read an env variable, so ... I guess ts is interpreted, so it doesn't matter; edit code makes the change happen immediately.

src/renderer/util/index.ts
```diff
@@ -41,26 +71,57 @@ export async function determineFilesFromNestedPaths(
...
+  // otherwise written like: /.zarr|.sldy$/i
+  const matcher = new RegExp(combinedExtensions + "$", 'i');
+
+  // If the regex matches it will return an array (truthy).
+  // If the regex doesn't match it will return null (falsy).
+  return Boolean(filePath.match(matcher)) && USE_MULTIFILE_ASSUMPTION;
+}
+
+/**
+ * Returns the total size of a given directory's children, sub-children, etc.
```
kmitcham, line 99: What happens when following links?  possibly circular links?  Not a blocker, but definitely something to think about.

kmitcham (approved):

src/renderer/util/index.ts
tyler-foster: Yeah, good point. I'll take a quick peek at what it would take to default to `false` but configure the tests to run such that it's `true`.

src/renderer/util/index.ts
tyler-foster: Oh my. Links? Like symlinks? That's definitely not a use-case I was considering... Does FSS already account for that in some way on its end?

src/renderer/services/file-storage-service/index.ts
```diff
@@ -111,7 +112,8 @@ export default class FileStorageService extends HttpCacheClient {
       // so the conversion must be manual each request
       file_size: fileSize,
       local_nas_shortcut: localNasPath !== undefined,
-      local_nas_path: localNasPath
+      local_nas_path: localNasPath,
+      multifile: isMultifile !== undefined ? isMultifile : false,
```
SeanLeRoy, line 116:
```suggestion
      multifile :!!isMultifile,
```

SeanLeRoy:
In the UI will it just show one upload for a ZARR or will it show one for each subfile?

SeanLeRoy (approved):

src/renderer/state/upload/logics.ts
```diff
@@ -724,25 +730,29 @@ const uploadWithoutMetadataLogic = createLogic({
...
+                    originalPath: filePath,
+                    shouldBeInArchive: true,
+                    shouldBeInLocal: true,
+                  },
+                  microscopy: {},
+                },
+                user,
+                {
+                  groupId,
+                  multifile: isMultifile
```
hughes036, line 752: Does this undergo another transformation to create the request body?  Asking because I expected the  `multifile` arg to be at the same level as `shouldBeInArchive` and `shouldBeInLocal`

src/renderer/util/index.ts
```diff
@@ -33,6 +34,26 @@ const canUserRead = async (filePath: string): Promise<boolean> => {
...
+    return [path];
+  }
+  const canRead = await canUserRead(path);
+  if (!canRead) {
+    throw new Error(`User does not have permission to read ${path}`);
+  }
+  const pathsUnderFolder = await fsPromises.readdir(path, {
+    withFileTypes: true,
+  });
+  return pathsUnderFolder
```
hughes036, line 52: So if it is not a regular file or a multifile, return a list of every path in the dir, ignoring subdirs?

src/renderer/util/index.ts
```diff
@@ -41,26 +62,66 @@ export async function determineFilesFromNestedPaths(
...
+/**
+ * Use a given filepath's extension to determine if it is a "multifile".
+ * @param filePath Path to the file
+ */
+export function determineIsMultifile(filePath: string): boolean {
+  ///////////////////////////////////////////////////////////////////
+  // TODO - Remove this once multifile support is feature-complete //
+  //         Flip the boolean to "true" for testing                //
+  ///////////////////////////////////////////////////////////////////
+  const USE_MULTIFILE_ASSUMPTION = process.env.USE_MULTIFILE_ASSUMPTION !== undefined
```
hughes036, line 81: `process.env.USE_MULTIFILE_ASSUMPTION !== undefined && process.env.USE_MULTIFILE_ASSUMPTION` would be a little more concise, arguably less readable (and very Matty-esque)

src/renderer/util/index.ts
```diff
@@ -41,26 +62,66 @@ export async function determineFilesFromNestedPaths(
...
+      ? process.env.USE_MULTIFILE_ASSUMPTION === 'true'
+      : false;
+  ///////////////////////////////////////////////////////////////////
+
+  const multifileExtensions = ['.zarr', '.sldy'];
+  const combinedExtensions = multifileExtensions.join('|');
+
+  // "ends with one of the listed extensions, ignoring casing"
+  // otherwise written like: /.zarr|.sldy$/i
+  const matcher = new RegExp(combinedExtensions + "$", 'i');
```
hughes036, line 91: Maybe the matching should be a different unit, so that it can be tested separately from the env var flag.

src/renderer/util/index.ts
hughes036: Never thought of it either.  Worth testing, maybe blocking if it causes any problems.

src/renderer/util/index.ts
```diff
@@ -41,26 +62,66 @@ export async function determineFilesFromNestedPaths(
...
+ * Borrowed from StackOverflow: https://stackoverflow.com/a/69418940
+ */
+export async function getDirectorySize(dir: string): Promise<number> {
+  const files = await readdir(dir, { withFileTypes: true });
+
+  const paths = files.map(async file => {
+    const path = join(dir, file.name);
+
+    if (file.isDirectory()) {
+      return await getDirectorySize(path);
```
hughes036, line 112: What about the other contents of `dir`?  Won't this result in just returning the size of the first subdir encountered?  It seems like it should be creating a list of all the file sizes and recursive calls, and summing it all.

src/renderer/util/index.ts
kmitcham: Fss stores the subfiles in a list, and checks that entries aren't already in the list before adding them.

src/renderer/services/file-storage-service/index.ts
kmitcham: I'm usually more in favor of verbosity over cool code things, but JS isn't my bag so it isn't an informed opinion in this case.

hughes036: I would call this 'Matty-esque'

hughes036:
> In the UI will it just show one upload for a ZARR or will it show one for each subfile?

One for the .zarr.

src/renderer/util/index.ts
tyler-foster: Does this links question seem worth an issue?

src/renderer/services/file-storage-service/index.ts
tyler-foster: I say we try it out

src/renderer/state/upload/logics.ts
tyler-foster: The request body is constructed in the file-storage-service module. I honestly wasn't sure how this object was supposed to be structured and was vaguely aware that the `service_fields` were a grab-bag... Should I move it up to `file`?

src/renderer/util/index.ts
tyler-foster: To be 100% clear - I didn't write this exact function, I just extracted it  from the below function of similar name because I thought it would look cleaner, though that was ultimately irrelevant to what I ended up doing.

But to answer your question, yeah, that's the existing behavior. If it's not a regular file or a multifile, then it it's a directory, and (for some reason) we only grab the top-level files in a given directory and not any in its sub-dirs. IDK if we want that long-term, but it's what we have.

src/renderer/util/index.ts
tyler-foster: I think I'll leave it. Hope is I'll be removing this whole chunk soon-ish anyways, right?

src/renderer/util/index.ts
tyler-foster: I think I've got a few arguments against that:

* The env var flag is going away someday anyways
* There's not much going on in this function beyond the regex matching anyways, so any existing unit tests should cover the regex matching.
* This would be the only place where that method would get used - writing a separate test for it would basically be testing a private method, wouldn't it?

src/renderer/util/index.ts
```diff
@@ -33,6 +34,26 @@ const canUserRead = async (filePath: string): Promise<boolean> => {
   }
 };
 
+async function determineFilesFromNestedPath(
```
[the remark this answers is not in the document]
tyler-foster: Added in f3492bea511b447cc2d1ed12c4dec99ed903b336

src/renderer/services/file-storage-service/index.ts
tyler-foster: changed in 9d9d665b6c4bd85c1f53ba4c5a5af41a5eba2168

src/renderer/util/index.ts
kmitcham: Maybe a TODO comment in the code?  It shouldn't be blocker, though.

src/renderer/util/index.ts
hughes036: > The env var flag is going away someday anyways.

But the regex will remain, right?  Im just talking about testing the regex part.

> There's not much going on in this function beyond the regex matching, so any existing unit tests should cover it.
> This would be the only place where that method would get used - writing a separate test for it would basically be testing a private method, wouldn't it?

tyler-foster: (we talked this throguh and decided the existing unit tests are sufficient)

tyler-foster:
(rebased on main)

PR #71, opened 2024-04-11
[FUA-70] Fix DateTime picker

tyler-foster:
### Context

Antoine tried to use the "argolight optical control" template when uploading a file and ran into an issue where the "Imaging Date" annotation would not correctly populate its value in the upload grid after pressing "OK". We then confirmed that all of the annotations with the DateTime annotation type have the same issue.

### Bug
The `<DateEditor>` component is basically just a wrapper around an antd component (antd is a big library we use all over the FUA). The antd component is intended to be super flexible and support multiple types of datetime entries, from ranges to dates to datetimes, etc. Thing is, its behavior changes quite a bit depending on what props you feed it.

This issue arose because we passed a function to the component's `onOk` prop, but didn't properly consume the parameter that it passed in. Kinda like this:

```
// inside the antd component
const innerFunction = (event, callbackFn) => {
    callbackFn(event.target.value);
}

// in our code
const callbackFn = () => {
    console.log("I don't take any arguments lol");
} 
```

### Changes
* In `<DateEditor>`
   * `handleCommit()` is now `handleCommit(moment)`
   *  Inner value tracked in state is now a `moment` object rather than a `Date` object.

### Testing
Only tested manually, but I'm very open to any unit test ideas.

### DIFF
--- a/src/renderer/containers/Table/Editors/DateEditor/index.tsx
+++ b/src/renderer/containers/Table/Editors/DateEditor/index.tsx
@@ -1,6 +1,6 @@
 import { DeleteOutlined } from "@ant-design/icons";
 import { Button, DatePicker } from "antd";
-import moment from "moment";
+import moment, {Moment} from "moment";
 import React, { useState } from "react";
 import { ColumnInstance } from "react-table";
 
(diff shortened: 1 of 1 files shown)


src/renderer/containers/Table/Editors/DateEditor/index.tsx
```diff
@@ -41,7 +41,7 @@ export default function DateEditor({
         (e.relatedTarget.tagName !== "LI" ||
           e.relatedTarget.attributes.getNamedItem("role")?.value !== "button"))
     ) {
-      handleCommit();
+      handleCommit(value);
```
hughes036, line 44: Was the fix just to pass `value` to the callback, or was it also changing the `Date` type to `Moment`?

hughes036 (approved):
Nice job on the quick turn around!

aswallace (approved):

src/renderer/containers/Table/Editors/DateEditor/index.tsx
tyler-foster: The fix was to pass the value into the callback.

Changing the type to `Moment` just helped make that cleaner, because it enabled me to pass `value` as a prop to `handleCommit` without having to make `handleCommit` handle both `Date` and `Moment` parameters. And the type of `value` was already being changed back and forth from date to moment in a couple of spots anyways.

kmitcham (approved):

Issue #76, opened 2024-04-30
Dev: Documentation should elaborate on how JSS is used to monitor upload status


tyler-foster:
Somewhere in this repo we should write out an explanation of how an upload's status is tracked via the Job Status Service. Upload progress reporting is dependent on both the File Storage Service and the FMS Mongo ETL providing updates to the Job Status Service, while the Upload App is subscribed to those updates via JSS, but we have no written explanation anywhere that can act as a guide by a developer who's new to the repo.

Issue #75, opened 2024-04-30
Multifile Uploads / LocalNAS Uploads (maybe?) never get confirmed as ETL'd


tyler-foster:
There are multiple progress reporting stages for any given FUA upload:

Uploading (featuring a loading spinner and 3 substages) -> Upload completed but not yet ETL'd (green question mark icon) -> Completed and ETL'd (green checkmark icon).

For some reason, my multifile uploads, which do get FileIds that can be used to view the files via the File Explorer, never transition to the final "green checkmark" stage. I've had this issue with a few other local NAS uploads as well, but with non-multifile uploads it seems more like an intermittent issue.

Clearly the files ARE getting ETL'd, so there must be an issue where the job updates for multifile uploads are never seen by the File Upload App. This should be fixed so that our users can know when their uploads are viewable in the File Explorer.

pgarrison:
I think this is resolved

PR #77, opened 2024-04-30
[FUA-68] Display multifile upload progress

tyler-foster:
#### Context
I added basic support for multifiles in the file upload app [in this previous PR](https://github.com/aics-int/aics-file-upload-app/pull/69), but that PR did not include progress reporting for those files (which was dependent on an FSS change). This PR does! Now the little spinny thing on the lefthand side of the "Uploads" table will actually spin when uploading a multifile. Which is important because otherwise users would sit there getting no feedback.

#### Changes
* I found the overall job status updating code to be a bit hard to read, so I broke it off into its own module, `handleUploadJobUpdates.ts`.
  * This currently only contains one function, but theoretically we could have a module that contains ALL of the logic relating to JSS job updates, retries, etc. This one function was even already being exported just for testing purposes (which is smelly).
* Updated the `handleUploadJob` function to report on progress for multifile uploads.
* Added unit test
  * The `App.test.ts` file suspiciously only tests `handleUploadJobUpdates()`

#### Testing
Added a new unit test, and performed several test uploads with multifiles.

#### Notes / Additional work
* Multifile uploads, or at least mine, still never transition to the "green checkmark" stage in the File Upload App's status column. This would imply that they aren't getting ETL'd into the explorer's mongoDB database, but that doesn't seem to be the case.
   * Made [a new issue here](https://github.com/aics-int/aics-file-upload-app/issues/75)
* The way we use JSS across the entire upload process (in the File Upload App, File Storage Service, FMS Mongo ETL) should be documented somewhere.
   * Made [an issue here](https://github.com/aics-int/aics-file-upload-app/issues/76)

### DIFF
--- a/src/renderer/containers/App/handleUploadJobUpdates.ts
+++ b/src/renderer/containers/App/handleUploadJobUpdates.ts
@@ -1,0 +1,115 @@
+import { Dispatch } from "react";
+
+import { FSSUpload, UploadStatus } from "../../services/file-storage-service";
+import {
+  JSSJob,
+  Service,
+  UploadJob,
+} from "../../services/job-status-service/types";
+import {
+  receiveFSSJobCompletionUpdate,
+  receiveJobUpdate,
+  updateUploadProgressInfo,
+} from "../../state/job/actions";
+import { Step } from "../Table/CustomCells/StatusCell/Step";
+
+/**
+ * Handles job updates for a standard file upload.
+ * Reports progress on the pre-upload MD5 step, the upload step, and then the post-upload MD5 step.
+ */
+function handleFSSJobUpdate(job: FSSUpload, dispatch: Dispatch<any>) {
+  const totalBytes = job.serviceFields.fileSize || 0; // 0 is a safe default, but in practice filesize is initialized immediately after job creation.
+  if (
+    job.serviceFields?.preUploadMd5 &&
+    job.serviceFields.preUploadMd5 !== job.serviceFields.fileSize
+  ) {
+    dispatch(
+      updateUploadProgressInfo(job.jobId, {
+        bytesUploaded: job.serviceFields?.preUploadMd5,
+        totalBytes,
+        step: Step.ONE,
+      })
+    );
+  } else if (
+    job.serviceFields?.currentFileSize &&
+    job.serviceFields.currentFileSize !== job.serviceFields?.fileSize
+  ) {
+    dispatch(
+      updateUploadProgressInfo(job.jobId, {
+        bytesUploaded: job.serviceFields?.currentFileSize,
+        totalBytes,
+        step: Step.TWO,
+      })
+    );
+  } else if (
+    job.serviceFields?.postUploadMd5 &&
+    job.serviceFields.postUploadMd5 !== job.serviceFields?.fileSize
+  ) {
+    dispatch(
+      updateUploadProgressInfo(job.jobId, {
+        bytesUploaded: job.serviceFields?.postUploadMd5,
+        totalBytes,
+        step: Step.THREE,
+      })
+    );
+  }
+}
+
+/**
+ * Handles a job update for a multifile (.zarr, .sldy) upload.
+ * Jumps straight to step 3 of 3 and reports progress as the sum of all bytes uploaded for relevant subfiles divided by
+ *  the expected total size of the upload.
+ */
+function handleFSSMultifileJobUpdate(job: FSSUpload, dispatch: Dispatch<any>) {
+  if (job.serviceFields?.subfiles) {
+    const totalBytesUploaded: number = Object.values(
+      job.serviceFields.subfiles
+    ).reduce((accum: number, value: number) => accum + value, 0);
+    dispatch(
+      updateUploadProgressInfo(job.jobId, {
+        bytesUploaded: totalBytesUploaded,
+        totalBytes: job.serviceFields?.fileSize || 0,
+        step: Step.THREE,
+      })
+    );
+  }
+}
+
+/**
+ * Updates the upload progress UI when JSS jobs get updated.
+ *
+ * @param job JSSJob that has been recently updated.
+ * @param dispatch The Redux Dispatch function.
+ */
+export function handleUploadJobUpdates(job: JSSJob, dispatch: Dispatch<any>) {
+  if (job.service === Service.FILE_STORAGE_SERVICE) {
+    // An FSS job update happens when:
+    //   * fileId has been published
+    //   * progress has been published on a pre-upload md5, file upload, or post-upload md5
+    //   * progress has been published on a multifile upload's subfile
+    //   * the upload app has initialized a retry
+    const fssJob = job as FSSUpload;
+
+    // If a fileId is present, the upload has completed and should be marked as such.
+    // If the upload job has become inactive or requires a retry, mark it as "failed".
+    if (
+      job.serviceFields?.fileId ||
+      job.currentStage === UploadStatus.INACTIVE ||
+      job.currentStage === UploadStatus.RETRY
+    ) {
+      dispatch(receiveFSSJobCompletionUpdate(fssJob));
+    }
+
+    // Otherwise, report progress
+    // TODO Multifile FSS jobs are currently identifiable by their service_fields.subfiles property,
+    //       but more ideally would just be labeled with something like "multifile: true".
+    if (fssJob.serviceFields?.subfiles) {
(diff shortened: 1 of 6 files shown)


src/renderer/containers/App/handleUploadJobUpdates.ts
```diff
@@ -0,0 +1,115 @@
...
+  ) {
+    dispatch(
+      updateUploadProgressInfo(job.jobId, {
+        bytesUploaded: job.serviceFields?.postUploadMd5,
+        totalBytes,
+        step: Step.THREE,
+      })
+    );
+  }
+}
```
tyler-foster, line 56: `handleFSSJobUpdate()` is largely just copy-pasted from its original form in `App/index.tsx`, but I'm open to comments / suggestions related to it.

src/renderer/containers/App/test/App.test.ts
```diff
@@ -64,4 +64,35 @@ describe("App", () => {
...
+                    'fileid2': 10,
+                    'fileid3': 15
+                },
+            },
+        };
+        let actionPersisted = undefined;
+        const dispatch = (action: Action)=>{
+            actionPersisted = action;
+        };
+        const expectedAction = updateUploadProgressInfo(fssJob.jobId, { bytesUploaded: 35, totalBytes: fssJob.serviceFields?.fileSize, step: Step.THREE })
```
hughes036, line 92: Is step 3 the 'still uploading' step?

src/renderer/containers/App/handleUploadJobUpdates.ts
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Handles a job update for a multifile (.zarr, .sldy) upload.
+ * Jumps straight to step 3 of 3 and reports progress as the sum of all bytes uploaded for relevant subfiles divided by
+ *  the expected total size of the upload.
+ */
+function handleFSSMultifileJobUpdate(job: FSSUpload, dispatch: Dispatch<any>) {
```
hughes036, line 63: So this moved out of `App/index` , and was updated to deal with multifile specific fields right?

src/renderer/containers/App/handleUploadJobUpdates.ts
```diff
@@ -0,0 +1,115 @@
...
+ * @param job JSSJob that has been recently updated.
+ * @param dispatch The Redux Dispatch function.
+ */
+export function handleUploadJobUpdates(job: JSSJob, dispatch: Dispatch<any>) {
+  if (job.service === Service.FILE_STORAGE_SERVICE) {
+    // An FSS job update happens when:
+    //   * fileId has been published
+    //   * progress has been published on a pre-upload md5, file upload, or post-upload md5
+    //   * progress has been published on a multifile upload's subfile
+    //   * the upload app has initialized a retry
```
hughes036, line 90: great commenting

src/renderer/containers/App/handleUploadJobUpdates.ts
```diff
@@ -0,0 +1,115 @@
...
+    const fssJob = job as FSSUpload;
+
+    // If a fileId is present, the upload has completed and should be marked as such.
+    // If the upload job has become inactive or requires a retry, mark it as "failed".
+    if (
+      job.serviceFields?.fileId ||
+      job.currentStage === UploadStatus.INACTIVE ||
+      job.currentStage === UploadStatus.RETRY
+    ) {
+      dispatch(receiveFSSJobCompletionUpdate(fssJob));
```
hughes036, line 100: These statuses render differently though - how is that handled downstream?

hughes036 (approved):

src/renderer/containers/App/test/App.test.ts
tyler-foster: Step 3 is the "post-upload MD5" step, but I don't believe any of the steps are labeled in a way that the users could tell what's going on.

src/renderer/containers/App/handleUploadJobUpdates.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    // Otherwise, report progress
+    // TODO Multifile FSS jobs are currently identifiable by their service_fields.subfiles property,
+    //       but more ideally would just be labeled with something like "multifile: true".
```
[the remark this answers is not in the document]
tyler-foster: agreed, and I forgot to make a ticket for it

src/renderer/containers/App/handleUploadJobUpdates.ts
```diff
@@ -0,0 +1,115 @@
...
+  receiveJobUpdate,
+  updateUploadProgressInfo,
+} from "../../state/job/actions";
+import { Step } from "../Table/CustomCells/StatusCell/Step";
+
+/**
+ * Handles job updates for a standard file upload.
+ * Reports progress on the pre-upload MD5 step, the upload step, and then the post-upload MD5 step.
+ */
+function handleFSSJobUpdate(job: FSSUpload, dispatch: Dispatch<any>) {
```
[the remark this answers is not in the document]
tyler-foster: I don't think JavaScript supports it directly, but there's a convention that we've never really used which is prepending function names with an `_` like in Python.

src/renderer/containers/App/handleUploadJobUpdates.ts
tyler-foster: Yeah, honestly I find this confusing. It's just the existing logic for completing a job. But here, "complete" means something more like "stopped". [Here's the redux-logic that gets called from here](https://github.com/aics-int/aics-file-upload-app/blob/main/src/renderer/state/job/logics.ts#L170-L184). You can see that there's a step where the job gets failed if it's in the retry or inactive stage. Realistically this should probably be done in a different logic called `receiveFSSJobFailedUpdate`

src/renderer/containers/App/handleUploadJobUpdates.ts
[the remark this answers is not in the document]
tyler-foster: Made the issue for it here:
https://github.com/aics-int/file-storage-service/issues/309

src/renderer/containers/App/handleUploadJobUpdates.ts
```diff
@@ -0,0 +1,115 @@
...
+  const totalBytes = job.serviceFields.fileSize || 0; // 0 is a safe default, but in practice filesize is initialized immediately after job creation.
+  if (
+    job.serviceFields?.preUploadMd5 &&
+    job.serviceFields.preUploadMd5 !== job.serviceFields.fileSize
```
kmitcham, line 24: Why would the preUploadMd5 ever equal the fileSize?  When they are both default initialized to zero or something?

src/renderer/containers/App/handleUploadJobUpdates.ts
```diff
@@ -0,0 +1,115 @@
...
+    dispatch(
+      updateUploadProgressInfo(job.jobId, {
+        bytesUploaded: job.serviceFields?.currentFileSize,
+        totalBytes,
+        step: Step.TWO,
+      })
+    );
+  } else if (
+    job.serviceFields?.postUploadMd5 &&
+    job.serviceFields.postUploadMd5 !== job.serviceFields?.fileSize
```
kmitcham, line 46: So I'm guessing this is another 'check the defaults' condition?  Step.THREE seems to be 'while the post upload md5 verification is running'

kmitcham (approved):

src/renderer/containers/App/handleUploadJobUpdates.ts
tyler-foster: This is copy-pasted / moved code that is certainly not "self-documenting"... so I'm not sure.

See line 28 just below this where the value for `bytesUploaded` is the `preUploadMd5` value. Can't say I have any clue what that actually means. I would assume `preUploadMd5` also starts at 0, just like how `totalbytes` is defaulted to on line 21. So you may be right about this being a check for... reasons, at the start of each step, but again I'm not sure.

tyler-foster: Could warrant some investigation just for better understanding. Another reason why some kind of documentation re: how we use JSS in here could be useful

src/renderer/containers/App/handleUploadJobUpdates.ts
tyler-foster: Ultimately the same discussion here:
https://github.com/aics-int/aics-file-upload-app/pull/77#discussion_r1588025100

PR #78, opened 2024-05-07
Bugfix: Enable multifile boolean for templated uploads

tyler-foster:
#### Context
There are two separate codepaths for uploads in the File Upload App - one for uploads without metadata, and one for uploads with metadata. Somehow, [in my initial PR](https://github.com/aics-int/aics-file-upload-app/pull/69), I didn't update the codepath where we correctly label an upload as a multifile upload in the request to FSS for uploads that have metadata. Whoops! That's kind of important.

This was an easy change though. Just check out the diff in `src/renderer/state/upload/logics.ts`. The rest are test updates.

### DIFF
--- a/src/renderer/state/upload/logics.ts
+++ b/src/renderer/state/upload/logics.ts
@@ -197,9 +197,14 @@ const initiateUploadLogic = createLogic({
     let uploads: UploadJob[];
     try {
       uploads = await Promise.all(
-        requests.map((request) =>
-          fms.initiateUpload(request, user, { groupId })
-        )
+          requests.map((request) => {
+                const serviceFields = {
+                  groupId,
+                  multifile: determineIsMultifile(request.file.originalPath),
+                }
+                return fms.initiateUpload(request, user, serviceFields);
+              }
+          )
       );
     } catch (error) {
       dispatch(
--- a/src/renderer/state/upload/test/logics.test.ts
+++ b/src/renderer/state/upload/test/logics.test.ts
@@ -225,6 +225,46 @@ describe("Upload logics", () => {
       expect(groupIds).to.not.be.lengthOf(fms.initiateUpload.callCount);
     });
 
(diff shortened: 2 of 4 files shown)


src/renderer/state/upload/logics.ts
```diff
@@ -197,9 +197,14 @@ const initiateUploadLogic = createLogic({
...
       uploads = await Promise.all(
-        requests.map((request) =>
-          fms.initiateUpload(request, user, { groupId })
-        )
+          requests.map((request) => {
+                const serviceFields = {
+                  groupId,
+                  multifile: determineIsMultifile(request.file.originalPath),
+                }
+                return fms.initiateUpload(request, user, serviceFields);
```
SeanLeRoy, line 205: Should `fms.initiateUpload` just handle checking if this is `multifile`?

SeanLeRoy (approved):

src/renderer/state/upload/test/logics.test.ts
```diff
@@ -225,6 +225,46 @@ describe("Upload logics", () => {
...
+      expect(
+          actions.includesMatch({
+            autoSave: true,
+            type: INITIATE_UPLOAD,
+          })
+      ).to.be.true;
+      expect(actions.list.find((a) => a.type === INITIATE_UPLOAD_SUCCEEDED)).to
+          .not.be.undefined;
+
+      // Assert that each upload had the expected "multifile" value
```
kmitcham, line 257: So if someone else adds an item to the test list, this test needs updating?

kmitcham (approved):

hughes036 (approved):

src/renderer/state/upload/logics.ts
tyler-foster: I'll have to look at it again, but ideally yes! I'll do that as part of a separate ticket though.

Issue #74, opened 2024-04-23
Date/Datetime pickers display values in their UTC times


tyler-foster:
[Example 1](https://github.com/aics-int/aics-file-upload-app/blob/main/src/renderer/containers/Table/DefaultCells/DisplayCell/index.tsx#L48)

[Example 2](https://github.com/aics-int/aics-file-upload-app/blob/main/src/renderer/containers/Table/DefaultCells/DisplayCell/index.tsx#L52)

[Example 3](https://github.com/aics-int/aics-file-upload-app/blob/main/src/renderer/containers/Table/Editors/DateEditor/index.tsx#L57)

Is there some un-commented business reason for this? It doesn't make any sense to me. Users select a specific date and time e.g. `2024-01-01 12:00 PM` and it gets converted to `2024-01-01 20:00:00` or something like that. Seems strange. Could fix it in five seconds, but I'm not sure why this is the way it is to begin with.

PR #79, opened 2024-05-07
(Tiny) [#74] Stop displaying selected date / datetime values as their UTC equivalents

tyler-foster:
### Context / Reasoning
For some reason, the date + datetime annotation value components will display a user's selected value as its UTC equivalent. In other words, someone will mark that a file was generated at 12:00 PM our time, and then the Upload App UI will display that value as if the user had selected 7 or 8 PM instead. Pretty weird.

I think this change is worthwhile because a) the current functionality could be confusing (it's confusing to me!), and b) it doesn't seem like the values saved in state ever get converted to UTC time anywhere else in the app anyways - it's only when they're displayed as values in the upload table. Seems non-sensical, or maybe the result of some decision from years past? We could have at least left a comment about it...

### DIFF
--- a/src/renderer/containers/Table/DefaultCells/DisplayCell/index.tsx
+++ b/src/renderer/containers/Table/DefaultCells/DisplayCell/index.tsx
@@ -45,11 +45,11 @@ export const useDisplayValue = (value?: ColumnValue, type?: ColumnType) =>
         return value[0] ? "Yes" : "No";
       case ColumnType.DATE:
         return (value as Date[])
-          .map((d) => moment(d).utc().format("M/D/YYYY"))
+          .map((d) => moment(d).format("M/D/YYYY"))
           .join(", ");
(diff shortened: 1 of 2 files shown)


kmitcham (approved):
Hmm.  I wish we knew why the UTC was there in the first place.

SeanLeRoy (approved):
Weird... I assume you double checked this gets read in as the expected date when viewing uploaded files but just in case not I'd try that out too

saeliddp (approved):

tyler-foster:
> Weird... I assume you double checked this gets read in as the expected date when viewing uploaded files but just in case not I'd try that out too

@SeanLeRoy Good point, I guess that could be the reasoning. I haven't checked yet but that's probably what's going on. Will have to look into updating things on that side.

hughes036 (approved):
Unit test?

PR #82, opened 2024-05-15
Bugfix: Fix multifile upload retries, progress reporting

tyler-foster:
#### Context
I recently tried uploading a `.sldy` sub-directory, and it seemed to be broken on two fronts:

1. If I closed the app mid-upload and then re-opened it, progress reporting wouldn't start up again (progress for the file would say `Step 1 of 3: 0%`)
2. If I didn't close the app and just let the upload run to completion, it would seem to get stuck at `Step 3 of 3: 99%`.

#### Changes
Problem 1 seemed to be cropping up because files wouldn't properly be labeled as multifiles when re-registered. This would seem to be a result of the `service_fields` property for a resumed job not getting updated correctly somewhere in our retry code.

Problem 2 seemed to be a result of a control flow mistake in the `handleUploadJobUpdates` module. The module would dispatch `receiveFSSJobCompletionUpdate()` when a job was completed (good), but then if the `serviceFields.subfiles` property was still present, it would move into the block where it labels the progress percentage in step 3 of 3 (bad - it should be "done").

### DIFF
--- a/src/renderer/containers/App/handleUploadJobUpdates.ts
+++ b/src/renderer/containers/App/handleUploadJobUpdates.ts
@@ -98,15 +98,15 @@ export function handleUploadJobUpdates(job: JSSJob, dispatch: Dispatch<any>) {
       job.currentStage === UploadStatus.RETRY
     ) {
       dispatch(receiveFSSJobCompletionUpdate(fssJob));
-    }
-
-    // Otherwise, report progress
-    // TODO Multifile FSS jobs are currently identifiable by their service_fields.subfiles property,
-    //       but more ideally would just be labeled with something like "multifile: true".
-    if (fssJob.serviceFields?.subfiles) {
-      handleFSSMultifileJobUpdate(fssJob, dispatch);
     } else {
-      handleFSSJobUpdate(fssJob, dispatch);
+        // Otherwise, report progress
+        // TODO Multifile FSS jobs are currently identifiable by their service_fields.subfiles property,
+        //       but more ideally would just be labeled with something like "multifile: true".
+        if (fssJob.serviceFields?.subfiles) {
+            handleFSSMultifileJobUpdate(fssJob, dispatch);
+        } else {
+            handleFSSJobUpdate(fssJob, dispatch);
+        }
     }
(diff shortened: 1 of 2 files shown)


src/renderer/containers/App/handleUploadJobUpdates.ts
```diff
@@ -98,15 +98,15 @@ export function handleUploadJobUpdates(job: JSSJob, dispatch: Dispatch<any>) {
...
-    }
-
-    // Otherwise, report progress
-    // TODO Multifile FSS jobs are currently identifiable by their service_fields.subfiles property,
-    //       but more ideally would just be labeled with something like "multifile: true".
-    if (fssJob.serviceFields?.subfiles) {
-      handleFSSMultifileJobUpdate(fssJob, dispatch);
     } else {
-      handleFSSJobUpdate(fssJob, dispatch);
+        // Otherwise, report progress
```
hughes036, line 102: Weird diff - it looks like the change is only that the block got indented?

tyler-foster: Yeah the diff makes it hard to read. The change is that everything following line 100 is now in a separate `else`. So if the `jobs.serviceFields.fileId || ...` condition starting on line 96 is met, the rest of the code gets skipped.

src/renderer/services/file-management-system/index.ts
```diff
@@ -119,7 +119,7 @@ export default class FileManagementSystem {
     // Grab file details
     const source = upload.serviceFields.files[0]?.file.originalPath;
     const fileName = path.basename(source);
-    const isMultifile = upload.serviceFields.multifile || false;
+    const isMultifile = determineIsMultifile(upload.jobName);
```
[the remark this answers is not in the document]
tyler-foster: `determineIsMultifile` is the function that checks to see if the given file path or name ends in a `.sldy` or `.zarr` suffix. It's used in other places to set the `serviceFields.multifile` value.

kmitcham (approved):

hughes036 (approved):

PR #84, opened 2024-06-12
Stop using "electron-webpack"; Switch from Yarn to NPM

tyler-foster:
### Context
The File Upload App build stopped building. Why? Because sometimes it does that. Yeah, that's right, only sometimes. `ESOCKETTIMEDOUT`. That's what Yarn Classic (occasionally) throws out when it tries to install a really big package (e.g. `antd`). Look, [I'm not the only person who has run into this](https://github.com/yarnpkg/yarn/issues/8242). And [the solution is simple](https://github.com/yarnpkg/yarn/issues/8242#issuecomment-753473953). Okay, that's easy, the solution is to not use Yarn Classic.

So we use something else. Yarn is up to V4, so we could use that. But wait, Yarn 2+ requires Node 18+. Okay, that's easy, just use Node 18+.

So we use Node 18+. But wait. Using Node 18+ with Webpack <4 means your cryptographic hashes aren't supported anymore. `Error: error:0308010C:digital envelope routines::unsupported`. What does that mean? I'm not totally sure. [But you can overwrite it in your WebPack config](https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported). Okay, that's easy, just overwrite that setting in your Webpack config. Maybe just upgrade Webpack.

So we update our WebPack config. But wait. _We don't have a WebPack config_. We use a library for handling our WebPack builds, called `electron-webpack`. Okay. Is there a newer version of it that works with newer Node and Webpack versions? Nope. [It was last updated 4 years ago](https://www.npmjs.com/package/electron-webpack). Great.

So our main issue is that we use an old, unsupported library for our WebPack builds, and now we want to update WebPack. The real solution? Well, I figure we should at least stop using `electron-webpack`.

### Changes
I'll give `electron-webpack` this at least. It [offers a method of exposing what it does under the hood](https://webpack.electron.build/extending-as-a-library#configuration-files). You can run a command to print out the WebPack configs it creates. That's cool. So I used that to get a jumpstart on creating our new WebPack configs.

Then I copied a bunch of stuff from the File Explorer App, which never used `electron-webpack` itself. That was a big help - thanks Gabe.

So the main changes are:
* Dumped `electron-webpack` in favor of managing our WebPack configuration ourselves
   * There are two configs in an Electron app - a `main` config, for the main Electron process, and the `renderer` config, which handles the actual rendering of the UI.
   * I am not super well-versed in manual WebPack configuration.
   * I may have gotten several things wrong.
   * Fonts aren't loading properly, and probably a few other style things I haven't noticed yet.
* Started using NPM instead of Yarn
   * I don't believe this is wildly impactful. We should be able to switch back to Yarn pretty easily if we choose to do so.
* Upgraded to WebPack v5.
* Pinned all the existing `package.json` dependencies, because some of them were giving me grief.
* Copy-pasted the `start-dev.js` script from the File Explorer App.

### Testing
I have:
* Managed to [get the release action to run to completion and produce files in S3](https://github.com/aics-int/aics-file-upload-app/actions/runs/9471926339). These appear to work correctly. The UI at least renders across Windows / Mac / Ubuntu. I checked.
* Successfully uploaded files from the MacOS build.

I have NOT:
* Successfully uploaded (read: tried to run) a multifile upload, which I'll need `/allen` drive access for. I'll test it on my Ubuntu setup soon.

### Reviewing
I would [appreciate it if you downloaded your OS' version of the v3.2.2-snapshot File Upload App build from S3](https://us-west-2.console.aws.amazon.com/s3/buckets/file-upload-app.allencell.org?region=us-west-2&bucketType=general&tab=objects) and then tried uploading a file. Otherwise, if you're a WebPack guru, check out the `webpack.<>.config.js` files in the diff for me and tell me if they make any sense.

### TODO
* ~~Figure out why fonts don't load right.~~ [Made a ticket.](https://github.com/aics-int/aics-file-upload-app/issues/85)
* ~~Make the build produce artifacts that match the format expected by the download page, or vice versa.~~
* Attempt a multifile upload

### Notes
I experimented a bit with trying to re-create the build using `electron-forge`, which is like `electron-webpack` and `electron-builder`, but like, officially supported by the main Electron people. I think. I [made a repo for that here](https://github.com/aics-int/aics-file-upload-app-electron-forge), but haven't gotten too deep into that process. `electron-forge` clearly gets a lot of support, so it seems like it could be worth pivoting to if it simplifies the build process enough.


scripts/start-dev.js
tyler-foster, line 1: Blatantly stolen from [here](https://github.com/AllenInstitute/aics-fms-file-explorer-app/blob/main/packages/desktop/scripts/start-dev.js).

package.json
```diff
@@ -46,7 +53,8 @@
     "react",
     "aics"
   ],
-  "author": "Lisa Harrylock",
```
tyler-foster, line 49: image

aswallace:
> I would [appreciate it if you downloaded your OS' version of the v3.2.2-snapshot File Upload App build from S3](https://us-west-2.console.aws.amazon.com/s3/buckets/file-upload-app.allencell.org?region=us-west-2&bucketType=general&tab=objects) and then tried uploading a file.

Was mostly able to do this on Mac with the 3.2.2-snapshot.0-arm64 version, although not until running `xattr -c` on the app. Currently unable to install via aics/ on Mac for 3.2.0 and above, getting an `AccessDenied` error

scripts/start-dev.js
```diff
@@ -0,0 +1,104 @@
+/**
+ * The purpose of this script is to coordinate the start up of all of the processes necessary for development
+ * in the context of Electron. This includes:
+ *   - webpack-dev-server, which watches, recompiles, and serves code run in Electron's renderer process
+ *   - webpack, which compiles and watches the code run in Electron's main process
+ *   - electron itself, which should only be started when the others are up and running
+ *
+ * This script is intended to be run in the foreground. Sending a SIGINT to this script will clean up the child
+ * processes spawned herein.
+ */
```
kmitcham, line 10: Maybe add a comment with a pointer to the original, in case we find a bug here and want to keep them both correct?

src/main/index.ts
```diff
@@ -51,7 +50,7 @@ function createMainWindow() {
           }
 
           mainWindow
-              .loadURL(`http://localhost:${process.env.ELECTRON_WEBPACK_WDS_PORT}`)
+              .loadURL(`http://localhost:8080`) // todo pull from constants file
```
kmitcham, line 53: checking intentional TODO left undone?

src/renderer/containers/CustomDataTable/test/selectors.test.ts
kmitcham, line 1: Not a fan of commenting out a whole file.  At least, add a comment at top explaining how/why it is happening.

kmitcham:
Holding off to approve until I get back in the office tomorrow and can test more usefully

src/renderer/containers/CustomDataTable/test/selectors.test.ts
tyler-foster: TBH I totally forgot I did this and I don't know how I missed it looking back at the diff.

webpack/webpack.renderer.config.js
```diff
@@ -0,0 +1,263 @@
...
+
+const { devServer } = require("./constants");
+
+module.exports = ({ production })  => {
+  const mode = production ? 'production' : 'development';
+
+  return {
+    context: path.resolve(__dirname, '..'),
+    devtool: 'eval-source-map',
+    externals:
```
pgarrison, line 17: Ideally we would do some DRY between this file and `webpack.renderer.js` with a shared `webpack.common.js`

pgarrison (approved):
Did not run but code looks good to me

scripts/start-dev.js
tyler-foster: Done in f3089ea67

src/renderer/containers/CustomDataTable/test/selectors.test.ts
tyler-foster: Done in c9f312db8 , no fancy related issues, seems whatever led to me commenting it out was fixed by other stuff I did.

src/renderer/containers/TemplateEditorModal/AnnotationEditorModal/test/AnnotationEditorModal.test.tsx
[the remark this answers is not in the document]
tyler-foster: Done in here 5ef89f32bc98 , featuring a note about a dumb bug and now a linting error that I need to go fix.

hughes036:
Looks like a Windows artifact was not created? 
image

scripts/start-dev.js
```diff
@@ -0,0 +1,110 @@
+/**
+ * Copied and pasted from the File Explorer App:
+ * https://github.com/AllenInstitute/aics-fms-file-explorer-app/blob/main/packages/desktop/scripts/start-dev.js
```
hughes036, line 3: good comment

hughes036 (approved):

hughes036:
Icon looks weird in the draw n drop dialog 
image

hughes036:
looks normal from launchpad though 
image

hughes036:
hm, somehow the old version and this one both can exist in my /Applications
image

hughes036:
launch pad is not showing the new version, and when i try to execute it manually it shows
image

@aswallace is this the one you downloaded?
image

tyler-foster:
> Looks like a Windows artifact was not created?

This is the goofy thing about the Windows builds. Your search is off by a hyphen. Check this out, excuse the different version numbers (doesn't matter):
image

tyler-foster:
> hm, somehow the old version and this one both can exist in my /Applications

@hughes036 This is because the name is changed to `File Upload App` from `file-upload-app` because of a line I added in `package.json`. I honestly might just change it back.

> launch pad is not showing the new version, and when i try to execute it manually it shows

Yup. The build isn't producing a DMG made for Macs that still have Intel processors. Kevin had the same issue. The Mac build only works with Apple Silicon at the moment :/

tyler-foster:
> > Looks like a Windows artifact was not created?
> 
> @hughes036 This is the goofy thing about the Windows builds. Your search is off by a hyphen.

Should be fixed now in ac2f31c65c0e816dc73a7b952f71339407a6e981 . IDK why I didn't do this 3 years ago.

tyler-foster:
> > hm, somehow the old version and this one both can exist in my /Applications
> 
> @hughes036 This is because the name is changed to `File Upload App` from `file-upload-app`.I honestly might just change it back.

Said "screw it" and added this back in b2307e782412d0d68db0a2780b4c8dd3ab067f8f. Thanks to another recent change it shouldn't get in the way of the artifact naming / download page functionality.


> > launch pad is not showing the new version, and when i try to execute it manually it shows
> 
> Yup. The build isn't producing a DMG made for Macs that still have Intel processors. Kevin had the same issue. The Mac build only works with Apple Silicon at the moment :/. So it works for me and Anya.

Should be fixed in 12ec5f0edbbd23c9abe2f7e648f7d998483cf10d. Which makes "universal" MacOS builds.

webpack/webpack.renderer.config.js
tyler-foster: Should be simple, but I made a separate ticket for it + will make a separate PR
https://github.com/aics-int/aics-file-upload-app/issues/86

src/main/index.ts
tyler-foster: Addressed in db8d371e09a2aecefe1ec07fefe59f8d1f668779 . Also highlighted the fact that one of my WebPack loaders was not in package.json 😬

tyler-foster:
PR is getting real big. Merging while acknowledging two new follow-up issues:
* https://github.com/aics-int/aics-file-upload-app/issues/85
* https://github.com/aics-int/aics-file-upload-app/issues/86

PR #87, opened 2024-06-17
[FUA-85] Load fonts correctly; Remove url-loader dependency

### LINKED ISSUE #85: Fonts don't load correctly following departure from electron-webpack
Fonts are not loading correctly in several parts of the app. For example, when hovering the "Upload" button, we'll see a serif font rather than a sans-serif font:

image

My first guess is that this is an issue with how fonts are loaded rather than an issue with how styles are loaded, but I could be wrong.

tyler-foster:
### Changes
* Fixes #85 
* Removes now-unnecessary `url-loader` dependency

Not sure _exactly_ why fonts weren't loading correctly following the switch I made to webpack v5, but v5 [has this cool new feature that simplifies asset loading and fixes this issue for us](https://webpack.js.org/guides/asset-modules/). It also means we don't need `url-loader` anymore - builds seem fine without it!

Compare this new pic:
image

To the pic from the ticket:
image

### DIFF
--- a/webpack/webpack.renderer.config.js
+++ b/webpack/webpack.renderer.config.js
@@ -132,19 +132,6 @@ module.exports = ({ production })  => {
                   'sass-loader']
             },
             {
-              test: /\.(png|jpe?g|gif)(\?.*)?$/,
-              use:
-                {
-                  loader: 'url-loader',
-                  options: {limit: 10240, name: 'imgs/[name]--[folder].[ext]'}
-                }
-            },
-            {
-              test: /\.(mp4|webm|ogg|mp3|wav|flac|aac)(\?.*)?$/,
-              loader: 'url-loader',
-              options: {limit: 10240, name: 'media/[name]--[folder].[ext]'}
-            },
-            {
               test: /\.ttf/,
               type: 'asset/resource'
             },
--- a/webpack/webpack.renderer.config.js
+++ b/webpack/webpack.renderer.config.js
@@ -145,12 +145,8 @@ module.exports = ({ production })  => {
               options: {limit: 10240, name: 'media/[name]--[folder].[ext]'}
             },
             {
-              test: /\.(woff2?|eot|ttf|otf)(\?.*)?$/,
-              use:
-                {
-                  loader: 'url-loader',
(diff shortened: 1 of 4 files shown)


webpack/webpack.renderer.config.js
```diff
@@ -132,25 +132,8 @@ module.exports = ({ production })  => {
...
-              options: {limit: 10240, name: 'media/[name]--[folder].[ext]'}
-            },
-            {
-              test: /\.(woff2?|eot|ttf|otf)(\?.*)?$/,
-              use:
-                {
-                  loader: 'url-loader',
-                  options: {limit: 10240, name: 'fonts/[name]--[folder].[ext]'}
-                }
+              test: /\.ttf/,
```
kmitcham, line 135: I don't understand this; we remove the special cases for a many extensions, but only have a special case left for ttf?  Is that the only extension that needs handling now?  wav|flac|png|woff2 all work fine without checks, but ttf still needs help?

tyler-foster: `wav|flac|png|woff2` _would not_ work without checks, but we don't use them. The only one of those extensions we're actually using in the repo is `ttf`.

The idea is to keep the config file as simple as possible. And if you consider the config file as a type of documentation in and of itself, these `module.rules` entries denote what to expect to find in the repository.

hughes036: So it relates to 'asset' (fonts / icons) being loaded, and the only assets that we use are fonts?

hughes036: how were they actually loaded?  was a URL to some file in the repo constructed?  What happens differently now?

hughes036 (approved):

webpack/webpack.renderer.config.js
tyler-foster: TTF = "TrueType font" = font files. We have one non-standard font we load into the app, _Nunito_, which was not being loaded correctly.

I'm not entirely sure how asset loading worked / works... The big difference now is that webpack does all of the work for us and we no longer need the special loader plugins.

kmitcham (approved):

PR #90, opened 2024-06-18
Update "Husky"

tyler-foster:
[Husky](https://typicode.github.io/husky/) allows us to set up and automatically run pre-commit hooks to make sure our code is clean and working and happy before we make commits. We already had this installed in the Upload App repo - this PR just updates the package and makes sure it still does what we want.

### DIFF
--- a/.huskyrc
+++ b/.huskyrc
@@ -1,3 +1,0 @@
-"hooks": {
-    "pre-commit": "lint-staged"
-}
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,2 @@
+npm run lint
+npm run test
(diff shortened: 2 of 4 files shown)


kmitcham (approved):

.huskyrc
```diff
@@ -1,3 +0,0 @@
-"hooks": {
-    "pre-commit": "lint-staged"
```
tyler-foster, line 2: `lint-staged` is actually another library that makes the linter only run on "staged" files. I need to get that to work again too, alongside `Prettier` 🤔

tyler-foster:
Going to close and later re-open + re-brand this PR with more commit hook add-ons.

PR #116, opened 2024-10-03
feature/toggle-local-upload

### LINKED ISSUE #115: Surface `should_be_in_local` for new uploads
**Objective**:

Surface pre-existing ability for FSS to do `cloud only` uploads

**Backend Design:**

It already exists!  Clients just need to surface the`shouldBeInLocal` arg of the existing Upload `/register` endpoint:

```jsx
POST /upload/register
{
	should_be_in_local: True | False // Note this can be camel or snake case.
}
```

Try it [[here](http://stg-aics-api.corp.alleninstitute.org/fss2/swagger-ui.html)

**Frontend Design:**

A UI control for choosing between a `cloudOnly` upload and a traditional `hybridStorage` upload.

For phase 1 of this feature, `should_be_in_local` will default to `False`.  See the roll out plan [here](https://www.notion.so/FMS-as-a-tool-for-reducing-on-prem-storage-footprint-for-AICS-7404373daf734e8d81aa7afc8ce049fd)

BrianWhitneyAI:
## Description 
The purpose of this PR is to surface the ability to toggle `shouldBeInLocal` (which is currently hard coded to `true`) so that users can upload to cloud only. This is part of a proof of concept that we want to implement as part of [this document](https://www.notion.so/FMS-as-a-tool-for-reducing-on-prem-storage-footprint-for-AICS-7404373daf734e8d81aa7afc8ce049fd).

Currently this toggle only effects the initial upload request and is only relevant to the upload with metadata option.

### Related Issues 
Resolves #115 

## Testing 

I have only tested this by debugging the process and looking at the file metadata. I may need assistance to test further (or determine what types of tests to write)  since I am new to this repo.

### DIFF
--- a/src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
+++ b/src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
@@ -5,8 +5,7 @@ import * as React from "react";
 import { useDispatch, useSelector } from "react-redux";
 
 import { closeUpload } from "../../../state/route/actions";
-import { getSelectedUploads } from "../../../state/selection/selectors";
-import { getShouldStoreLocally } from "../../../state/setting/selectors";
+import { getSelectedUploads, getShouldStoreLocally } from "../../../state/selection/selectors";
 import {
   initiateUpload,
   submitFileMetadataUpdate,
@@ -79,13 +78,14 @@ export default function PageFooter(props: Props) {
           "Upload"
         )}
       </Button>
-      <Checkbox
-        checked={shouldStoreLocally}
-        onChange={onCheckboxChange} // Update Redux state on change
-        style={{ marginLeft: 16 }}
-      >
-        Store Locally
-      </Checkbox>
+      <div className={styles.checkboxContainer}>
+        <Checkbox
+          checked={shouldStoreLocally}
+          onChange={onCheckboxChange}
+        >
+          Store Locally
+        </Checkbox>
+      </div>
     </div>
   );
 }
--- a/src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
+++ b/src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
@@ -5,11 +5,11 @@ import * as React from "react";
 import { useDispatch, useSelector } from "react-redux";
 
 import { closeUpload } from "../../../state/route/actions";
-import { getSelectedUploads, getShouldStoreLocally } from "../../../state/selection/selectors";
+import { getSelectedUploads, getShouldBeInLocal } from "../../../state/selection/selectors";
 import {
   initiateUpload,
   submitFileMetadataUpdate,
-  setShouldStoreLocally,
+  setShouldBeInLocal,
 } from "../../../state/upload/actions";
 import {
   getUploadValidationErrors,
@@ -33,7 +33,7 @@ export default function PageFooter(props: Props) {
   const selectedUploads = useSelector(getSelectedUploads);
   const isUploadInProgress = useSelector(getIsUploadInProgress);
   const validationErrors = useSelector(getUploadValidationErrors);
-  const shouldStoreLocally = useSelector(getShouldStoreLocally);
+  const ShouldBeInLocal = useSelector(getShouldBeInLocal);
 
   function onSubmit() {
     props.onSubmit();
@@ -47,9 +47,9 @@ export default function PageFooter(props: Props) {
     }
   }
 
-  // Handler to update shouldStoreLocally state when the checkbox is toggled
+  // Handler to update ShouldBeInLocal state when the checkbox is toggled
   const onCheckboxChange = (e: CheckboxChangeEvent) => {
-    dispatch(setShouldStoreLocally(e.target.checked));
+    dispatch(setShouldBeInLocal(e.target.checked));
   };
 
   return (
@@ -80,7 +80,7 @@ export default function PageFooter(props: Props) {
       </Button>
       <div className={styles.checkboxContainer}>
         <Checkbox
-          checked={shouldStoreLocally}
+          checked={ShouldBeInLocal}
           onChange={onCheckboxChange}
         >
           Store Locally
--- a/src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
+++ b/src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
@@ -6,6 +6,7 @@ import { useDispatch, useSelector } from "react-redux";
 
 import { closeUpload } from "../../../state/route/actions";
 import { getSelectedUploads } from "../../../state/selection/selectors";
+import { getShouldStoreLocally } from "../../../state/setting/selectors";
 import {
   initiateUpload,
   submitFileMetadataUpdate,
@@ -14,7 +15,6 @@ import {
 import {
   getUploadValidationErrors,
 } from "../../../state/upload/selectors";
-import { getShouldStoreLocally } from "../../../state/setting/selectors";
 import { getCanSubmitUpload, getIsUploadInProgress } from "../selectors";
 
 const styles = require("./styles.pcss");
--- a/src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
+++ b/src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
@@ -1,5 +1,6 @@
 import { LoadingOutlined } from "@ant-design/icons";
-import { Button } from "antd";
+import { Button, Checkbox } from "antd";
+import { CheckboxChangeEvent } from "antd/es/checkbox";
 import * as React from "react";
 import { useDispatch, useSelector } from "react-redux";
 
@@ -8,8 +9,12 @@ import { getSelectedUploads } from "../../../state/selection/selectors";
 import {
   initiateUpload,
   submitFileMetadataUpdate,
+  setShouldStoreLocally,
 } from "../../../state/upload/actions";
-import { getUploadValidationErrors } from "../../../state/upload/selectors";
+import {
+  getUploadValidationErrors,
+} from "../../../state/upload/selectors";
+import { getShouldStoreLocally } from "../../../state/setting/selectors";
 import { getCanSubmitUpload, getIsUploadInProgress } from "../selectors";
 
 const styles = require("./styles.pcss");
@@ -29,6 +34,7 @@ export default function PageFooter(props: Props) {
   const selectedUploads = useSelector(getSelectedUploads);
   const isUploadInProgress = useSelector(getIsUploadInProgress);
   const validationErrors = useSelector(getUploadValidationErrors);
+  const shouldStoreLocally = useSelector(getShouldStoreLocally);
 
   function onSubmit() {
     props.onSubmit();
@@ -41,6 +47,11 @@ export default function PageFooter(props: Props) {
       }
     }
   }
+
+  // Handler to update shouldStoreLocally state when the checkbox is toggled
+  const onCheckboxChange = (e: CheckboxChangeEvent) => {
+    dispatch(setShouldStoreLocally(e.target.checked));
+  };
 
   return (
     <div className={styles.pageFooter}>
@@ -68,6 +79,13 @@ export default function PageFooter(props: Props) {
           "Upload"
         )}
       </Button>
+      <Checkbox
+        checked={shouldStoreLocally}
+        onChange={onCheckboxChange} // Update Redux state on change
+        style={{ marginLeft: 16 }}
+      >
+        Store Locally
+      </Checkbox>
     </div>
   );
 }
--- a/src/renderer/services/file-management-system/index.ts
+++ b/src/renderer/services/file-management-system/index.ts
@@ -225,7 +225,7 @@ export default class FileManagementSystem {
       };
       await this.mms.createFileMetadata(fileId, metadataWithUploadId);
   
-      const { localPath, cloudPath } = await this.fss.getFileAttributes(fileId);
+      const { localPath, cloudPath, name } = await this.fss.getFileAttributes(fileId);
       const readPath = localPath ?? cloudPath;
       await this.jss.updateJob(
         upload.jobId,
@@ -235,7 +235,7 @@ export default class FileManagementSystem {
             result: [
               {
                 fileId,
-                fileName: path.basename(metadata.file.originalPath), 
+                fileName: name, 
                 readPath: readPath,
               },
             ],
--- a/src/renderer/services/file-management-system/index.ts
+++ b/src/renderer/services/file-management-system/index.ts
@@ -236,7 +236,7 @@ export default class FileManagementSystem {
               {
                 fileId,
                 fileName: name, 
-                readPath: readPath,
+                readPath,
               },
             ],
           },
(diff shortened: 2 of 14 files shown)


BrianWhitneyAI:
This is what the toggle looks like Screenshot 2024-10-03 at 3 16 15 PM

BrianWhitneyAI:
With box clicked 
Screenshot 2024-10-03 at 3 32 52 PM

BrianWhitneyAI:
With box unclicked
Screenshot 2024-10-03 at 3 33 28 PM

src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
```diff
@@ -68,6 +79,13 @@ export default function PageFooter(props: Props) {
           "Upload"
         )}
       </Button>
+      <Checkbox
```
SeanLeRoy, line 82: This should be centered within the container (the checkbox should also be centered to the text)

src/renderer/state/upload/selectors.ts
```diff
@@ -460,7 +462,7 @@ export const getUploadRequests = createSelector(
           FileType.OTHER,
         originalPath: filePath,
         shouldBeInArchive: true,
-        shouldBeInLocal: true,
+        shouldBeInLocal: shouldStoreLocally ?? true,
```
SeanLeRoy, line 465: shouldStoreLocally should have a default so it is never `null` or `undefined`

src/renderer/state/setting/reducer.ts
```diff
@@ -28,6 +28,7 @@ export const initialState: SettingStateBranch = {
     [AlertType.INFO]: true,
     [AlertType.DRAFT_SAVED]: false,
   },
+  shouldStoreLocally: true,
```
SeanLeRoy, line 31: I think this state field makes more sense in the `selection` state branch IMO
https://github.com/aics-int/aics-file-upload-app/blob/a48c70336d0e3d4bb7c8146854e43338482846b9/src/renderer/state/selection/reducer.ts

src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
```diff
@@ -1,15 +1,20 @@
 import { LoadingOutlined } from "@ant-design/icons";
-import { Button } from "antd";
+import { Button, Checkbox } from "antd";
+import { CheckboxChangeEvent } from "antd/es/checkbox";
 import * as React from "react";
 import { useDispatch, useSelector } from "react-redux";
 
 import { closeUpload } from "../../../state/route/actions";
 import { getSelectedUploads } from "../../../state/selection/selectors";
+import { getShouldStoreLocally } from "../../../state/setting/selectors";
```
hughes036, line 9: `shouldBeInLocal` is not the greatest name choice, but I think it makes sense to be consistent and use that name rather than adding a new name (`getShouldStoreLocally`) which represents the same info

hughes036 (commented):
Looks good! Sounds like you have not tested against stg yet, want to pair on that?

src/renderer/state/upload/selectors.ts
BrianWhitneyAI: resolved in [4e933f6](https://github.com/aics-int/aics-file-upload-app/pull/116/commits/4e933f6acba2f622da8d9fe9a79869cc552ee5a0)

src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
BrianWhitneyAI: resolved in [4e933f6](https://github.com/aics-int/aics-file-upload-app/pull/116/commits/4e933f6acba2f622da8d9fe9a79869cc552ee5a0)

src/renderer/state/setting/reducer.ts
BrianWhitneyAI: resolved in [4e933f6](https://github.com/aics-int/aics-file-upload-app/pull/116/commits/4e933f6acba2f622da8d9fe9a79869cc552ee5a0)

src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
BrianWhitneyAI: I dont know if I agree with this. The state that we are defining is that new uploads should store locally meaning that it should pass true to new calls that involve shouldBeInLocal, but isnt the exact flag itsself.

SeanLeRoy (approved):

BrianWhitneyAI:
@hughes036 @saeliddp I added the register functionality you were talking about. Some notable things came out of it. When uploading with should_be_in_local a local path is not created. This means that when we try to fetch this in the complete function [here](https://github.com/aics-int/aics-file-upload-app/blob/88508b9f2318add7e6c35553e11d9fbd64ff69ad/src/renderer/services/file-management-system/index.ts#L227), we do not report anything. I added some logic [here](https://github.com/aics-int/aics-file-upload-app/blob/1110ac3428703e48e395a27ff93a49b3b9cb346a/src/renderer/services/file-management-system/index.ts#L229) to default to the cloud path if that does not exist which I think will work for now but will likely need to be updated.



Example file of a upload with should_be_in_local = false 
```
{
  "fileId": "bd11ccb83ea84dc3a9012fe425ae5501",
  "name": "test6.csv",
  "type": "csv",
  "size": 341,
  "md5": "5688a2a31e2c381e3f9677a73942117f",
  "cloudPath": "staging.files.allencell.org/501/ae5/425/2fe/901/c3a/84d/3ea/cb8/11c/bd/test6.csv",
  "localPath": null,
  "created": 1728424157978,
  "file_id": "bd11ccb83ea84dc3a9012fe425ae5501",
  "cloud_path": "staging.files.allencell.org/501/ae5/425/2fe/901/c3a/84d/3ea/cb8/11c/bd/test6.csv",
  "local_path": null
}
```

And what that result looks like currently as a completed upload (Note the cloud path)
Screenshot 2024-10-08 at 2 51 26 PM

src/renderer/services/file-management-system/index.ts
```diff
@@ -233,8 +235,8 @@ export default class FileManagementSystem {
             result: [
               {
                 fileId,
-                fileName: path.basename(localPath),
-                readPath: localPath,
+                fileName: path.basename(metadata.file.originalPath),
```
hughes036, line 238: is `originalPath` the path to the file on the user's system that FUA is uploading?  If so, the file can be given a different name in FMS using the `fileName` arg or the `/register` endpoint.  I would stick to basing this `fileName` key on `readPath` if that works.

src/renderer/services/file-storage-service/test/file-storage-service.test.ts
```diff
@@ -49,7 +49,8 @@ describe("FileStorageService", () => {
         file_size: fileSize,
         local_nas_path: undefined,
         local_nas_shortcut: false,
-        multifile: false
+        multifile: false,
+        should_be_in_local: undefined
```
hughes036, line 53: Shouldn't `should_be_in_local` always get a boolean value ?

BrianWhitneyAI: should_be_in_local is optional right now and since we are simulating a call we can choose what output we want that to be. We could add should_be_in_local to the required variables and choose to inject the value here. However this doesn't change any of the test logic.

src/renderer/services/file-management-system/index.ts
BrianWhitneyAI: resolved in  [80a6911](https://github.com/aics-int/aics-file-upload-app/pull/116/commits/80a6911252d78d883e9b898ff9055f72500ae3ea) by just using the name from the file attires. Don't know the implications of this usage though.

src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
BrianWhitneyAI: The distinction I was making was:

shouldStoreLocally = store all new uploads made by app locally 

versus the specific tag

should_be_in_local  which is the metadata value

I honestly dont have too strong of an opinion and am happy to change it. just like this message if I should!

src/renderer/services/file-storage-service/test/file-storage-service.test.ts
BrianWhitneyAI: Thinking about it a bit more I think it is actually better to have it set as `undefined`. The variable is passed through the registerUpload function. At a higher level we set this value but no changes occur within the registerUpload function itself, by setting it to undefined we can see more easily if some new modification shows up ie defaulting to true or false. Which before would have been invisible.

saeliddp: > should_be_in_local is optional right now

Optional in the UI as well as the function? If it's optional in the UI, I agree it should be optional in the function and tested as `undefined` (as well as t/f). However, if it's always defined in the UI, then it seems like it should not be optional in the function or need to be tested as `undefined`

saeliddp (approved):
My review should be weighted lightly since I have little react/js experience, but looks good!

saeliddp:
> When uploading with should_be_in_local a local path is not created

If `should_be_in_local` is false, we do not expect a local path to be created. I'm not super familiar with JSS, but the fact that the attribute is called `readPath` seems to suggest that any readable path should be allowed as the value there? In which case, it would make sense to me to always use the cloud path since we intend to always have files in the cloud, but they will only sometimes be in local.

src/renderer/services/file-storage-service/test/file-storage-service.test.ts
saeliddp: Update: it looks like it should always be defined in the UI since it's a checkbox, so I'm not sure I understand your second comment--maybe we can discuss at an FMS standup?

src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
hughes036: > store all new uploads made by app locally

Sorry this doesnt clear it up for me.  

Doesn't 'store new upload locally' mean the same thing as 'set `should_be_in_local` equal to true for the upload' ?  

Im not trying to be difficult, I really am trying to understand the distinction.

If they mean something different, I can understand a different variable name, although it will be confusing to have 2 variables with such similar names.  If it really is distinct, I would suggest making the name less similar to the existing variable name.

src/renderer/state/setting/reducer.ts
hughes036: I think thats a safer way to do it.

src/renderer/services/file-management-system/index.ts
hughes036: i think this is the safer thing to do.
Comparing a JSS job for a new upload to an older one would confirm.

src/renderer/services/file-storage-service/test/file-storage-service.test.ts
hughes036: I notice that `local_nas_path` is also undefined also.  So I probably am just not understanding this test very well.  But since all the other args are ones that would be valid on the server, I think it would be easier to understand if that was made consistent.

hughes036:
> @hughes036 @saeliddp I added the register functionality you were talking about. Some notable things came out of it. When uploading with should_be_in_local a local path is not created. This means that when we try to fetch this in the complete function [here](https://github.com/aics-int/aics-file-upload-app/blob/88508b9f2318add7e6c35553e11d9fbd64ff69ad/src/renderer/services/file-management-system/index.ts#L227), we do not report anything. I added some logic [here](https://github.com/aics-int/aics-file-upload-app/blob/1110ac3428703e48e395a27ff93a49b3b9cb346a/src/renderer/services/file-management-system/index.ts#L229) to default to the cloud path if that does not exist which I think will work for now but will likely need to be updated.
> 
> Example file of a upload with should_be_in_local = false
> 
> ```
> {
>   "fileId": "bd11ccb83ea84dc3a9012fe425ae5501",
>   "name": "test6.csv",
>   "type": "csv",
>   "size": 341,
>   "md5": "5688a2a31e2c381e3f9677a73942117f",
>   "cloudPath": "staging.files.allencell.org/501/ae5/425/2fe/901/c3a/84d/3ea/cb8/11c/bd/test6.csv",
>   "localPath": null,
>   "created": 1728424157978,
>   "file_id": "bd11ccb83ea84dc3a9012fe425ae5501",
>   "cloud_path": "staging.files.allencell.org/501/ae5/425/2fe/901/c3a/84d/3ea/cb8/11c/bd/test6.csv",
>   "local_path": null
> }
> ```
> 
> And what that result looks like currently as a completed upload (Note the cloud path) Screenshot 2024-10-08 at 2 51 26 PM

I think we need to find out how the JSS `readPath` field is used.  

@tyler-foster insert distributed monolith critique here.

hughes036:
> `readPath` seems to suggest that any readable path should be allowed as the value there? In which case, it would make sense to me to always use the cloud path since we intend to always have files in the cloud, but they will only sometimes be in local.

Except that up until now, the localPath would always be a readable `/allen` path, and JSS clients may be using it with that expectation.  

Lets discuss @ standup.

src/renderer/services/file-management-system/index.ts
```diff
@@ -233,8 +235,8 @@ export default class FileManagementSystem {
             result: [
               {
                 fileId,
-                fileName: path.basename(localPath),
-                readPath: localPath,
+                fileName: name, 
+                readPath: readPath,
```
SeanLeRoy, line 239: LGTM! The result path here I think is just used for reading back to display to the user in the upload tracking UI as the path to go look for the file if they want to double check that it is uploaded.

Minor nit where if the key and value are the same you can just toss the variable in there
```suggestion
                readPath,
```

SeanLeRoy (approved):

src/renderer/services/file-storage-service/test/file-storage-service.test.ts
SeanLeRoy: This is just testing the FSS call not the data flow from the user's click to the server so it is undefined here because we aren't passing a value explicitly to FSS. IMO it doesn't matter whether you do or don't set it here as long as you assert for the expected value which you do on line 61 & 63

src/renderer/containers/UploadWithTemplatePage/PageFooter/index.tsx
SeanLeRoy: The distinction I think Brian is trying to make is that this flag is like a setting rather than a direct selection of the toggle. I think either are fine for this since it isn't common for server variable names to directly map to client state variable names in the app.

hughes036 (approved):

src/renderer/services/file-storage-service/test/file-storage-service.test.ts
tyler-foster: I'm afraid I still don't quite understand.

> by setting it to undefined we can see more easily if some new modification shows up ie defaulting to true or false

Is setting `should_be_in_local` to `undefined` a way to guard against any future changes that might change the value somewhere below this function? Is that the idea?

BrianWhitneyAI: yeah its the same way we handle local_nas_path [here](https://github.com/aics-int/aics-file-upload-app/blob/88508b9f2318add7e6c35553e11d9fbd64ff69ad/src/renderer/services/file-storage-service/test/file-storage-service.test.ts#L50)

PR #117, opened 2024-10-16
feature/multifile-id-job-updates

BrianWhitneyAI:
### Description 
The purpose of this PR is to resolve [TODO ](https://github.com/aics-int/aics-file-upload-app/blob/main/src/renderer/containers/App/handleUploadJobUpdates.ts#L103-L106) and use `multifile` service field in place of existing logic.


BrianWhitneyAI:
I am having a lot of trouble uploading zarrs to test this addition. The relevant tests seem to indicate correctness but want to do a full end to end test. Hoping to talk about it at standup.

PR #118, opened 2024-10-17
feature/multifile-id-job-updates

BrianWhitneyAI:
### Description 
The purpose of this PR is to resolve [TODO ](https://github.com/aics-int/aics-file-upload-app/blob/main/src/renderer/containers/App/handleUploadJobUpdates.ts#L103-L106) and use `multifile` service field in place of existing logic.

### DIFF
--- a/src/renderer/containers/App/handleUploadJobUpdates.ts
+++ b/src/renderer/containers/App/handleUploadJobUpdates.ts
@@ -73,6 +73,7 @@ function handleFSSMultifileJobUpdate(job: FSSUpload, dispatch: Dispatch<any>) {
       })
     );
   }
+  // TODO?: Maybe raise error here if there is no subfiles?
 }
 
 /**
@@ -100,9 +101,7 @@ export function handleUploadJobUpdates(job: JSSJob, dispatch: Dispatch<any>) {
       dispatch(receiveFSSJobCompletionUpdate(fssJob));
     } else {
         // Otherwise, report progress
-        // TODO Multifile FSS jobs are currently identifiable by their service_fields.subfiles property,
-        //       but more ideally would just be labeled with something like "multifile: true".
-        if (fssJob.serviceFields?.subfiles) {
+        if (fssJob.serviceFields?.multifile) {
             handleFSSMultifileJobUpdate(fssJob, dispatch);
         } else {
             handleFSSJobUpdate(fssJob, dispatch);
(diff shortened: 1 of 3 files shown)


src/renderer/containers/App/handleUploadJobUpdates.ts
```diff
@@ -73,6 +73,7 @@ function handleFSSMultifileJobUpdate(job: FSSUpload, dispatch: Dispatch<any>) {
       })
     );
   }
+  // TODO?: Maybe raise error here if there is no subfiles?
```
saeliddp, line 76: Good point. When FSS first creates an upload tracking job, there is no `serviceFields.subfiles` property. While I'm not sure exactly when in the upload FUA will get the first job update, it's possible that the very first update of a valid multifile upload will have no subfiles.

saeliddp (approved):

src/renderer/containers/App/handleUploadJobUpdates.ts
```diff
@@ -100,9 +101,7 @@ export function handleUploadJobUpdates(job: JSSJob, dispatch: Dispatch<any>) {
       dispatch(receiveFSSJobCompletionUpdate(fssJob));
     } else {
         // Otherwise, report progress
-        // TODO Multifile FSS jobs are currently identifiable by their service_fields.subfiles property,
-        //       but more ideally would just be labeled with something like "multifile: true".
-        if (fssJob.serviceFields?.subfiles) {
+        if (fssJob.serviceFields?.multifile) {
```
tyler-foster, line 104: Since this field ended up being `multifile` instead of `isMultifile`, maybe we should explicitly check this as `fssJob.serviceFields?.multifile === true`?

src/renderer/containers/App/handleUploadJobUpdates.ts
tyler-foster: Yeah, good point. There should be at least one place around here where the `subfiles` property is used to do some math - let's make sure there's a check, wherever that happens, for the existence of the `subfiles` property to begin with.

BrianWhitneyAI: We look at it [here ](https://github.com/aics-int/aics-file-upload-app/blob/734c6b07a5baa3e7c53f99d177ae067045f1074b/src/renderer/containers/App/handleUploadJobUpdates.ts#L64)

tyler-foster (approved):

SeanLeRoy (approved):

PR #130, opened 2025-01-29
(Small) [FUA-92] Change default environment to production

### LINKED ISSUE #92: New FUA versions default to "localhost" environment
~~For _devs_, the File Upload App should default to either a localhost or staging environment.~~ Separate ticket, if we do it at all.

For _users_, the File Upload App should default to pointing at production. Currently, the latest artifacts on the FUA download page (3.3.0+) default to localhost for users, which is not what we want.

tyler-foster:
closes #92 

### Description
* Changes the default target environment to `aics.corp.alleninstitute.org:80` instead of `localhost:8080`.

### Context
Without this, new Upload App users will default to looking at an error message saying `Failed to retrieve metadata: Request failed with status code 404`. Then the loading spinner on the landing page will spin forever. Won't make a great impression if we anticipate increased FUA usage following the move to cloud-first FMS!

### Recreating

See `electron-store` NPM page here:
https://www.npmjs.com/package/electron-store/v/5.2.0

The package we use for storing user configs places those config files under `app.getPath('userData')`. For MacOS, this comes out to `/Users/<your user>/Library/Application Support/file-upload-app`. Navigate to that path and delete `config.json` to reset your File Upload App config. Doing so before this branch has been merged will prove that the app defaults to the `localhost` environment setting, at least when launching in dev mode.

### Reviewing
One main file here, then just updated tests.
* [src/shared/constants.ts](https://github.com/aics-int/aics-file-upload-app/compare/feature/FUA-92_default_to_production?expand=1#diff-84451d0e7a53cf99f9eb786c5cebf4a96c2447f149c1ea0eece1f42c445b8eef)

### DIFF
--- a/src/renderer/services/http-cache-client/test/http-cache-client.test.ts
+++ b/src/renderer/services/http-cache-client/test/http-cache-client.test.ts
@@ -120,7 +120,7 @@ describe("HttpCacheClient", () => {
     await httpCacheClient.get(url);
     // eslint-disable-next-line @typescript-eslint/unbound-method
     expect(httpClient.get).to.have.been.calledOnceWith(
-      "http://localhost:8080/foo"
+      "http://aics.corp.alleninstitute.org:80/foo"
     );
   });
 });
--- a/src/renderer/state/setting/test/logics.test.ts
+++ b/src/renderer/state/setting/test/logics.test.ts
@@ -33,7 +33,7 @@ import {
 } from "../selectors";
 
 describe("Setting logics", () => {
-  const localhost = "localhost";
+  const productionHost = "aics.corp.alleninstitute.org";
   const stagingHost = "staging";
   let labkeyClient: SinonStubbedInstance<LabkeyClient>;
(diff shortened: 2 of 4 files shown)


hughes036 (approved):
Users of our release were defaulting to localhost?  Did you witness this in the wild?

tyler-foster:
> Users of our release were defaulting to localhost? Did you witness this in the wild?

I vaguely recall some instance of this happening in the wild, which should be why the issue exists in the first place, but I'm struggling to reproduce it at the moment because I can't remember how to reset my saved settings 🤔 .

pgarrison (changes requested):
Please wait on #139 before merging to preserve accurate commit history

tyler-foster:
> > Users of our release were defaulting to localhost? Did you witness this in the wild?
> 
> I vaguely recall some instance of this happening in the wild, which should be why the issue exists in the first place, but I'm struggling to reproduce it at the moment because I can't remember how to reset my saved settings 🤔 .

Added a section to this PR description for recreating the issue, for posterity. Merging now.

PR #158, opened 2025-02-07
(Tiny)[FUA-150] Change "Multifile" to "Chunked File"

### LINKED ISSUE #150: Update "Multifile" nomenclature to "Chunked Files"
Travis spoke to leadership about it, and apparently "Chunked File" is the term we're landing on for what today we call "Multifile". We should update the UI accordingly.

tyler-foster:
closes #150 

@TravisKroeker spoke with leadership, and apparently "chunked file" is the new term they landed on for what we now call "multifile". I'm not a huge fan of it personally... but it's not my decision! There are only a handful of places in the UI where we actually use the term, so the idea with this PR is that we can just change the appropriate `enum UploadType` value and use that value for referring to next generation distributed file formats in the UI wherever we do.

In the rest of the code it's still just "multifile" for now.

### DIFF
--- a/src/renderer/containers/UploadSelectionPage/UploadTypeSelector/index.tsx
+++ b/src/renderer/containers/UploadSelectionPage/UploadTypeSelector/index.tsx
@@ -37,7 +37,7 @@ export default function UploadTypeSelector() {
                     <div className={styles.radioButtonContent}>
                         <FileOutlined className={styles.radioButtonIcon} />
                         <div className={styles.radioButtonText}>
-                            <span className={styles.radioButtonTitle}>File</span>
+                            <span className={styles.radioButtonTitle}>{UploadType.File}</span>
                             <br />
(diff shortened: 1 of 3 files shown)


kmitcham (approved):
Does BFF have flag for this that needs to get updated?

SeanLeRoy (approved):
I think there is a real term that the zarr devs coined we can use and it isn't either of these, but I also don't remember it so 👍 works for me for now

tyler-foster:
> I think there is a real term that the zarr devs coined we can use and it isn't either of these, but I also don't remember it so 👍 works for me for now

IIRC Travis said that this was Next Generation File Format, like OME-NGFF, but we didn't want to use it because they've already got dibs (so it may not obviously apply to e.g. Nikon formats) and it won't age well.

BrianWhitneyAI (approved):

hughes036 (approved):

PR #160, opened 2025-02-07
[FUA-156] Add additional help text under "Upload Type" selector

### LINKED ISSUE #156: Add Upload Type selection help text
While there's already help text beneath the buttons for choosing an Upload Type on the Upload Selection page, we could use more to clarify our new functionality. Yeah, it's a bit ugly to double up on it, but it should be useful.

Follow designs here:
https://www.figma.com/design/uOgQ3a9MUk2BCQrEAfw8jZ/File-Upload-App?node-id=2101-9663&t=jaxKU8849LhnITSV-0

Reqs:
* For "File" uploads: `Select one or more standalone files. Attempting to select Chunked Files or folders will result in an error.`
* For "Chunked File" uploads: `Select one or more Chunked File folders. Each folder in your selection will be uploaded as a single FMS record. `

Copy may be adjusted according to user feedback...

tyler-foster:
closes #156

While the "Upload Type" radio buttons have help text on them already, Travis and I figured that the selection is still complicated enough to warrant additional hints as to what they do. This PR adds more help text that changes depending on the user's selected UploadType.

Before:
upload-type-before

After:

https://github.com/user-attachments/assets/43478db6-ad5b-406d-8a86-423059cb2757

### DIFF
--- a/src/renderer/containers/UploadSelectionPage/UploadTypeSelector/index.tsx
+++ b/src/renderer/containers/UploadSelectionPage/UploadTypeSelector/index.tsx
@@ -19,6 +19,12 @@ export default function UploadTypeSelector() {
         dispatch(selectUploadType(e.target.value));
     }
 
+    // Help text below the radio group. Defaults to the text for UploadType.File
+    let additionalHelpText = `Select one or more standalone files. Attempting to select ${UploadType.Multifile}s or folders will result in an error.`
+    if (uploadType === UploadType.Multifile) {
+        additionalHelpText = `Select one or more ${UploadType.Multifile} folders. Each folder in your selection will be uploaded as a single FMS record.`
+    }
+
     return (
         <div>
             <h1>
@@ -58,6 +64,11 @@ export default function UploadTypeSelector() {
                     </div>
                 </Radio.Button>
             </Radio.Group>
+            {uploadType !== null &&
+                <div className={styles.additionalHelpText}>
+                    {additionalHelpText}
+                </div>
+            }
         </div>
     )
 }
--- a/src/renderer/containers/UploadSelectionPage/UploadTypeSelector/styles.pcss
+++ b/src/renderer/containers/UploadSelectionPage/UploadTypeSelector/styles.pcss
@@ -1,3 +1,8 @@
+.additional-help-text {
+  padding-left: 5px;
+  padding-top: 5px;
+}
+
 .radio-button {
   align-items: center;
   display: flex;


src/renderer/containers/UploadSelectionPage/DragAndDropPrompt/styles.pcss
```diff
@@ -4,7 +4,7 @@
   display: flex;
   flex-direction: column;
   margin-bottom: 20px;
-  margin-top: 25px;
+  margin-top: 15px;
```
tyler-foster, line 7: Shortened this to account for the additional space created by the new <div> in `UploadTypeSelector`

src/renderer/containers/UploadSelectionPage/UploadTypeSelector/index.tsx
```diff
@@ -19,6 +19,12 @@ export default function UploadTypeSelector() {
         dispatch(selectUploadType(e.target.value));
     }
 
+    // Help text below the radio group. Defaults to the text for UploadType.File
+    let additionalHelpText = `Select one or more standalone files. Attempting to select ${UploadType.Multifile}s or folders will result in an error.`
+    if (uploadType === UploadType.Multifile) {
```
kmitcham, line 24: Isnt multifile getting renamed 'chunked file' by the higher-ups?

kmitcham:
The screen shot looks like it says multifile, and I thought I saw another PR saying we were changing the user-facing version of those sorts of files to 'Chunked File'

src/renderer/containers/UploadSelectionPage/UploadTypeSelector/index.tsx
hughes036: Is it possible to disable folders from selection?

hughes036 (approved):

src/renderer/containers/UploadSelectionPage/UploadTypeSelector/index.tsx
tyler-foster: > Isnt multifile getting renamed 'chunked file' by the higher-ups? AH, the enum is probably updated. Never mind.

Yeah, should get caught by the enum change.

 > Is it possible to disable folders from selection?

Yup! That's what it does now. [Electron makes it very easy](https://github.com/aics-int/aics-file-upload-app/blob/main/src/renderer/containers/UploadSelectionPage/index.tsx#L35-L45).

tyler-foster:
> The screen shot looks like it says multifile, and I thought I saw another PR saying we were changing the user-facing version of those sorts of files to 'Chunked File'

Should be handled by the enum change in a different PR, but I'll double check when I rebase in a minute here.

TravisKroeker (commented):
When "File" is selected, I would simplify the text of "Attempting to select..." to just "Selecting...". I know there is a technicality here about what the system will actually allow but I don't think we need to surface that nuance to the users. Keep it simple and consistent.

tyler-foster:
> When "File" is selected, I would simplify the text of "Attempting to select..." to just "Selecting...". I know there is a technicality here about what the system will actually allow but I don't think we need to surface that nuance to the users. Keep it simple and consistent.

Agreed. Changed in 5ba2ce967491e2db5db392d4e2867d212705430d

saeliddp (approved):

Issue #144, opened 2025-02-05
Dev: Replace Enzyme with Testing Library


tyler-foster:
As a dev, I want to manage as few dependencies as possible, and I want those dependencies to be well-maintained and work well with each other.

I've recently encountered some errors in trying to run tests here in the Upload App after upgrading `electron`. Those errors have been downstream of `enzyme`, which is a DOM testing tool that we use for ~5 tests and could easily replace with React Testing Library.

PR #161, opened 2025-02-07
[FUA-155] Move and restyle "Add More Files" button on "Add Metadata" page

### LINKED ISSUE #155: Change metadata selection page's "Add More Files" button
The button currently sits in an awkward spot between "cancel" and "upload", and it doesn't look much like a back button.
See design here:
https://www.figma.com/design/uOgQ3a9MUk2BCQrEAfw8jZ/File-Upload-App?node-id=2101-9663&t=jaxKU8849LhnITSV-0

Reqs:
* Move the button to the top-left of the page, in a fixed(ish) position.
* Border should appear when hovered to indicate it's a button.
* Add a left-arrow icon.
* Change copy to "Add More Files to Template" (or equivalent).

tyler-foster:
closes #155 

Since the upload process is now split across two pages, we had to add a "back" button on the second page. In our first design, it was sandwiched right between "cancel upload" and "submit", which was pretty ugly. This PR moves it up to the top left of the page and adds an arrow to suggest it'll return you to the previous page.

Before:
add_more_before

After:
add_more_after

(I am aware the table looks weird)

### DIFF
--- a/src/renderer/containers/AddMetadataPage/index.tsx
+++ b/src/renderer/containers/AddMetadataPage/index.tsx
@@ -1,3 +1,4 @@
+import { ArrowLeftOutlined } from "@ant-design/icons";
 import { Alert, Button, Spin } from "antd";
 import { ipcRenderer } from "electron";
 import React from 'react';
@@ -12,6 +13,7 @@ import {
   getUploadError,
 } from "../../state/feedback/selectors";
 import { getImagingSessions } from "../../state/metadata/selectors";
+import { selectPage } from "../../state/route/actions";
 import {
   getAreSelectedUploadsInFlight,
   getSelectedUploads,
@@ -25,8 +27,6 @@ import {
 import CustomDataTable from '../CustomDataTable';
 
 import AddMetadataPageFooter from './AddMetadataPageFooter';
-import { selectPage } from "../../state/route/actions";
-import { ArrowLeftOutlined } from "@ant-design/icons";
 
 const styles = require("./styles.pcss");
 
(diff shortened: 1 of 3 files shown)


src/renderer/containers/AddMetadataPage/index.tsx
```diff
@@ -72,66 +76,77 @@ export default function AddMetadataPage() {
...
+            {!selectedUploads.length && // If we're adding new files, not editing ones that have been uploaded.
+              <Button
+                className={styles.returnButton}
+                onClick={() => dispatch(selectPage(Page.UploadWithTemplate))}
+                type="ghost"
+              >
+                <span className={styles.returnButtonIcon}><ArrowLeftOutlined /></span> Add More Files to Template
+              </Button>
+            }
+            </div>
```
tyler-foster, line 90: Bit of a GitHub miss here - the above section is all that was added to this page. The rest was wrapped in another `<div>` and indented a bit.

kmitcham (approved):

src/renderer/containers/AddMetadataPage/index.tsx
```diff
@@ -72,66 +76,77 @@ export default function AddMetadataPage() {
...
-            </>
-            )}
-            {isSelectedJobLoading ? (
-            <div className={styles.spinContainer}>
-                <div>Loading...</div>
-                <Spin />
+        <div>
+            <div>
+            {!selectedUploads.length && // If we're adding new files, not editing ones that have been uploaded.
+              <Button
```
hughes036, line 82: thoughts on testing a that the DOM matches state?  I thought React makes this sort of thing pretty convenient?

hughes036 (approved):

src/renderer/containers/AddMetadataPage/index.tsx
tyler-foster: You're right, I should add that. Would be extra nice if I could get [this ticket to change which library we use for DOM testing](https://github.com/aics-int/aics-file-upload-app/issues/144) done at some point too.

TravisKroeker (approved):
As long as the hover state is as discussed it looks good to me!

PR #162, opened 2025-02-10
[FUA-148] Fix issue where non-zarr, non-sldy multifiles wouldn't upload properly.

### LINKED ISSUE #148: BUG: Multifile uploads still don't work for files not ending in `.zarr` or `.sldy`
The utility function `determineIsMultifile()` is still used in multiple places, and it still relies on uploaded multifiles ending with the `.zarr` or `.sldy` extensions. This is preventing us from properly uploading multifiles that don't end in those extensions.

Thanks to some relatively recent changes we've made, we should be able to determine whether a job belongs to a multifile solely based off of its `multifile: true` service field.

This may require either:
* Removing metadata-less uploads entirely.
* Restricting metadata-less uploads to just files, rather than folders.

Due to the fact that there's no way to properly identify a file vs. multifile upload without additional user input (like the "Upload Type" selection in the metadata w/ template flow).

tyler-foster:
closes #148 

### Context

While doing some manual testing, I discovered that multifile uploads that didn't end in `.zarr` or `.sldy` were failing to upload properly.

### Changes
* Uploads are now registered as "multifile" uploads [if their created JSS job has `service_fields.multifile: true`](https://github.com/aics-int/aics-file-upload-app/blob/main/src/renderer/state/upload/logics.ts#L204).
* Removed the `determineIsMultifile()` utility function that was previously used to figure out if an uploaded path was a multifile.
* Removed the `determineFilesFromNestedPath()` utility function that was previously used to handle "Folder" uploads.
* Users may no longer upload folders / multifiles without metadata.
   * We shouldn't let them upload _anything_ without metadata, tbh, but I'm calling that out of scope.

The existing tests for the `initiateUpload()` logic should still be sufficient here.


kmitcham (approved):

src/renderer/state/upload/logics.ts
```diff
@@ -732,12 +730,19 @@ const uploadWithoutMetadataLogic = createLogic({
...
-        deps.action.payload
-      );
+      // Don't let users upload folders / multifiles without metadata.
+      // This is mainly because there's no graceful way to handle different upload types with no "Upload Type" input,
+      //  but also because we want to discourage uploads without metadata in general (and this would allow them to
+      //  upload a lot of files w/o metadata all at once).
+      for (let i = 0; i < deps.action.payload.length; i += 1) {
+        const stats = await fs.promises.stat(deps.action.payload[i]);
+        if (stats.isDirectory()) {
+          throw Error(`Uploading folders without metadata is not allowed. Cannot upload "${deps.action.payload[i]}".`);
```
saeliddp, line 740: I might be getting my changes mixed up, but iirc, we aren't allowing non-multifile folder uploads now. If that's the case, should we replace 'folders' with 'multifiles' or 'chunked files'?

saeliddp (approved):

src/renderer/state/upload/logics.ts
tyler-foster: Yeah, fair question. It's pretty confusing. I should probably make it "folders and chunked files". The idea is that we should at least tell people that the old functionality w/ folders isn't available anymore.

tyler-foster: Changed here be8d32807d810656f817a0d473f6773df2f84d9e

src/renderer/state/upload/logics.ts
```diff
@@ -732,12 +730,19 @@ const uploadWithoutMetadataLogic = createLogic({
...
-        deps.action.payload
-      );
+      // Don't let users upload folders / multifiles without metadata.
+      // This is mainly because there's no graceful way to handle different upload types with no "Upload Type" input,
+      //  but also because we want to discourage uploads without metadata in general (and this would allow them to
+      //  upload a lot of files w/o metadata all at once).
+      for (let i = 0; i < deps.action.payload.length; i += 1) {
+        const stats = await fs.promises.stat(deps.action.payload[i]);
+        if (stats.isDirectory()) {
+          throw Error(`Uploading folders or ${UploadType.Multifile}s without metadata is not allowed. Cannot upload "${deps.action.payload[i]}".`);
```
hughes036, line 740: How do we know there is no metadata when this is reached?

> Uploading folders or
Isnt (non multifile) folders of files no longer a possibility?

src/renderer/state/upload/logics.ts
```diff
@@ -755,7 +760,7 @@ const uploadWithoutMetadataLogic = createLogic({
                 user,
                 {
                   groupId,
-                  multifile: isMultifile
+                  multifile: false // because we disallow uploading multifiles without metadata at all
```
hughes036, line 763: I agree that we should wholesale disallow uploads w/out metadata.  Can we identify a product owner / stakeholder list to dedide? @TravisKroeker ?

hughes036 (approved):

src/renderer/state/upload/logics.ts
tyler-foster: > How do we know there is no metadata when this is reached?
Easy, it's in the `uploadWithoutMetadata` function lol

> Isnt (non multifile) folders of files no longer a possibility?
Yeah, it's awkward, but part of the idea is to communicate that it's no longer possible.

PR #190, opened 2025-03-25
Fix for react devtools failing to load- ignore if it does

yrkim98:
Currently, when installing FUA with `npm install` react devtools will fail to load due to a bad version of `electron-devtools-instealler`. When react devtools fails to load, the whole electon app doesnt load as that will block app loading.

I've tried a couple different versions of `electron-devtools-installer` (`3.1.0`, `3.2.0`, and `4.0.0`) but couldnt resolve this issue. We should make it so that if react devtools fails to load- the rest of the app still loads correctly.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -61,8 +61,12 @@ function createMainWindow() {
                   console.error("Failed to load from webpack-dev-server", error);
               });
       })
-      .catch((err: Error) =>
-          console.error("An error occurred loading React Dev Tools: ", err)
+      .catch((err: Error) => {
+          console.log(`Failed to load React devtools \n ${err}`);
+          window.loadFile(path.join("dist", "renderer", "index.html")).catch((error: Error) => {
+            console.error("Failed to load from file", error);
+          });
+        }
       )
       .finally(() => window.webContents.openDevTools());
   } else {


tyler-foster (approved):

src/main/index.ts
```diff
@@ -61,8 +61,12 @@ function createMainWindow() {
                   console.error("Failed to load from webpack-dev-server", error);
               });
       })
-      .catch((err: Error) =>
-          console.error("An error occurred loading React Dev Tools: ", err)
+      .catch((err: Error) => {
+          console.log(`Failed to load React devtools \n ${err}`);
+          window.loadFile(path.join("dist", "renderer", "index.html")).catch((error: Error) => {
```
hughes036, line 66: What is going on in this fix?  How does it work?

hughes036 (commented):
Reviewers aside from Tyler and I aren't going to understand any context around this.  We want this app to have a broad dev base - can you describe the problem and add more devs?

hughes036 (approved):
> react devtools will fail to load due to a bad version of electron-devtools-instealler. 

 ⬆️ seems incongruous with ⬇️ 

> I've tried a couple different versions of electron-devtools-installer (3.1.0, 3.2.0, and 4.0.0) but couldnt resolve this issue.

If those versions all fail (with the same behavior?), then why do you conclude that the behavior is due to a bad version?

PR #191, opened 2025-03-27
Scheduled downtime dialog box

yrkim98:
When downtime is scheduled I am showing a error message dialog with js's `window.alert()`. Alternatively I could `dispatch(uploadFailed)` with a message indicating there is currently downtime scheduled, but no dialog box will show.

### DIFF
--- a/src/renderer/state/upload/logics.ts
+++ b/src/renderer/state/upload/logics.ts
@@ -226,7 +226,7 @@ const initiateUploadLogic = createLogic({
         await fms.upload(upload);
       } catch (error) {
         if (error?.response?.status === 503) {
-          window.alert("The upload service is undergoing scheduled maintenance (503). Please try again later.");
+          window.alert("The upload service is undergoing scheduled maintenance. Please try again later.");
           return;
         }
         dispatch(


hughes036 (approved):
This is good for a first pass, but is there an already established message dialog which matches the app look&feel that would be better?

Also, can you post a screenshot on the pr description?

src/renderer/state/upload/logics.ts
```diff
@@ -225,6 +225,10 @@ const initiateUploadLogic = createLogic({
       try {
         await fms.upload(upload);
       } catch (error) {
+        if (error?.response?.status === 503) {
+          window.alert("The upload service is undergoing scheduled maintenance (503). Please try again later.");
```
hughes036, line 229: I would leave the 503 out of the message - this is aimed a user who has no use for that detail.

tyler-foster, line 229: Instead of `window.alert()`, we should probably instead wrap this with `dispatch(uploadFailed())` like we see just below this. i haven't acutally checked, but I'm assuming that's the mechanism through which errors are displayed via the stylized error message pop-up rather than the alert dialogue.

hughes036: If this is not straightforward to get working, id be good with a release that uses window.alert, followed up by an improved/styled dialog.

tyler-foster (approved):

Issue #186, opened 2025-03-11
Make 'Program' a required annotation for any template.


hughes036:
AICS needs a way to generate reports (and eventually soft partitions) based cache usage by program.  The first step is for every FMS file to be annotated with a program.

PR #193, opened 2025-04-17
Make program a required annotation

yrkim98:
This implements https://github.com/aics-int/aics-file-upload-app/issues/186 which will make program a required annotation for all uploads / templates.

Currently, FUA enforces required annotations on the frontend by adding it to `PLATE_RELATED_COLUMNS` in `src/renderer/containers/CustomDataTable/selectors.tsx`. (This is how `Plate Barcode` and `Notes` behave as well), so, implemented this using the same pattern.

Changes:

- `src/renderer/constants/index.ts`:  Adds program as an `AnnotationName`
- `src/renderer/containers/CustomDataTable/selectors.tsx`: Adds a `ProgramCell` as one of the columns that show up by default with all templates in the annotation table during uploads
- `src/renderer/containers/Table/CustomCells/ProgramCell/index.tsx`: a `CustomCell` that shows a dropdown of all options for the `Program` annotation, and updates state based on user selection. This is based off of the [PlateBarcodeCell code ](https://github.com/aics-int/aics-file-upload-app/blob/main/src/renderer/containers/Table/CustomCells/PlateBarcodeCell/index.tsx), but is simplified to only be a dropdown (since users should not be able to define new programs, and only select from existing values).
- `src/renderer/containers/CustomDataTable/test/selectors.test.ts`: test fixes

Notes:
I will keep this a draft pr until i can test a full upload on staging. Need this approved: https://github.com/AllenCellSoftware/labkey-db-fixes/pull/144


src/renderer/constants/index.ts
```diff
@@ -10,6 +10,7 @@ export enum AnnotationName {
   IMAGING_SESSION = "Imaging Session",
```
hughes036, line 10: Imaging session is required too?

src/renderer/containers/CustomDataTable/test/selectors.test.ts
```diff
@@ -251,8 +251,8 @@ describe("CustomDataTable selectors", () => {
       const actual = getTemplateColumnsForTable(state);
 
       // Assert
-      expect(actual).to.be.lengthOf(6);
-      actual.slice(3).forEach((column, index) => {
+      expect(actual).to.be.lengthOf(7);
+      actual.slice(4).forEach((column, index) => { // 4 default "plate related" columns
```
hughes036, line 255: Im not sure what this comment is saying

src/renderer/containers/Table/CustomCells/ProgramCell/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+import { AnnotationName } from "../../../../constants";
+import { getAnnotations, getAnnotationOptions } from "../../../../state/metadata/selectors";
+import { FileModel } from "../../../../state/types";
+import { updateUpload } from "../../../../state/upload/actions";
+import DisplayCell from "../../DefaultCells/DisplayCell";
+
+const { Option } = Select;
+
+export default function ProgramCell(props: CellProps<FileModel, string[]>) {
```
hughes036, line 15: Is this following a pattern established for the other required annotations?

src/renderer/containers/Table/CustomCells/ProgramCell/styles.pcss
```diff
@@ -0,0 +1,17 @@
+.divider {
+    margin: 4px 0;
+}
+
+.menu {
```
hughes036, line 5: Was the extra col causing style problems?

hughes036 (approved):

src/renderer/containers/CustomDataTable/test/selectors.test.ts
hughes036: does that include imaging session?

saeliddp (approved):
Take my approval with a grain of salt since I have limited react/js experience--but it seems like you're following existing patterns here, so seems pretty safe!

src/renderer/containers/CustomDataTable/selectors.tsx
```diff
@@ -87,6 +88,16 @@ const DEFAULT_COLUMNS: CustomColumn[] = [
...
 
 export const PLATE_RELATED_COLUMNS: CustomColumn[] = [
+  {
+    accessor: AnnotationName.PROGRAM,
+    Cell: ProgramCell,
+    // This description was pulled from LK 04/17/25
+    description: "Name of project or program a dataset can be used for.",
+    width: getColumnWidthForType(
+      AnnotationName.PROGRAM,
+      ColumnType.LOOKUP
```
tyler-foster, line 98: I checked out this branch, and when I tried to select a value for the "Program" annotation (which does show up, which is good) all I got was an empty dropdown that said "no data". Have you managed to get the dropdown to populate properly?

For reference, here are cases where files are annotated with the "Program" annotation already:
https://stg-aics.corp.alleninstitute.org/labkey/query/AICS/executeQuery.view?schemaName=filemetadata&query.queryName=CustomAnnotationJunction&query.AnnotationId%2FName~contains=Program

PR #195, opened 2025-05-09
Fix bug with program dropdown

yrkim98:
This is part 2 of https://github.com/aics-int/aics-file-upload-app/pull/193, resolving a bug @tyler-foster found.

I made labkey db changes (on `stg`) to stop exposing the `Program` annotation to FUA so that it can't be manually added to templates. As a result the `Program` options were removed from the template's annotation list recieved from labkey -> dropdown for program was empty.

To fix I again treated this similarly to `PlateBarcode`. I am using `labkey-client` to get all the `Program` options and filling the dropdown with these values. This only happens when `ProgramCell` first renders.

New to redux, so could use a second look on how i'm using redux to fetch/store the `Program` options.

### DIFF
--- a/src/renderer/state/metadata/test/logics.test.ts
+++ b/src/renderer/state/metadata/test/logics.test.ts
@@ -15,6 +15,7 @@ import {
   LabkeyPlateResponse,
   LabkeyTemplate,
   ScalarType,
+  AnnotationOption,
 } from "../../../services/labkey-client/types";
 import MetadataManagementService from "../../../services/metadata-management-service";
 import { requestFailed } from "../../actions";
@@ -49,6 +50,8 @@ import {
   requestMetadata,
   requestTemplates,
   retrieveOptionsForLookup,
+  requestProgramOptions, 
+  receiveProgramOptions,
 } from "../actions";
 import { getBarcodeSearchResults } from "../selectors";
 
@@ -306,6 +309,31 @@ describe("Metadata logics", () => {
       // after
       expect(getBarcodeSearchResults(store.getState())).to.be.empty;
       expect(labkeyClient.getPlatesByBarcode.called).to.be.false;
+    });
+  });
+  describe("requestProgramOptions", () => {
+    it("dispatches receiveProgramOptions given OK response", async () => {
+      const fakeOptions: AnnotationOption[] = [
+        { annotationOptionId: 1, annotationId: 153, value: "Variance" },
+        { annotationOptionId: 2, annotationId: 153, value: "EMT" },
+      ];
+  
+      labkeyClient.getProgramOptions.resolves(fakeOptions);
+  
+      await runRequestSucceededTest(
+        requestProgramOptions(),
+        receiveProgramOptions(fakeOptions)
+      );
+    });
+  
+    it("dispatches requestFailed given error", async () => {
+      labkeyClient.getProgramOptions.rejects(new Error("backend boom"));
+  
+      await runRequestFailedTest(
+        requestProgramOptions(),
+        "Could not retrieve program options: backend boom",
+        AsyncRequest.GET_PROGRAM_OPTIONS
+      );
     });
   });
 });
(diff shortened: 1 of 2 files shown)


src/renderer/containers/Table/CustomCells/ProgramCell/index.tsx
```diff
@@ -1,34 +1,35 @@
...
 
 export default function ProgramCell(props: CellProps<FileModel, string[]>) {
   const dispatch = useDispatch();
 
-  const annotations = useSelector(getAnnotations);
-  const options = useSelector(getAnnotationOptions);
-
-  // Get options for program annotation
-  const programAnnotation = annotations.find(a => a.name === AnnotationName.PROGRAM);
-  const programOptions = programAnnotation
```
hughes036, line 23: Is this previous special treatment for `Program` from your earlier work?

src/renderer/containers/Table/CustomCells/ProgramCell/index.tsx
```diff
@@ -1,34 +1,35 @@
...
-    ? options.filter(opt => opt.annotationId === programAnnotation.annotationId)
-    : [];
+  const programOptions = useSelector(getProgramOptions);
```
hughes036, line 20: Looks like the logic for setting program options moved to a different Redux layer?

src/renderer/containers/Table/CustomCells/ProgramCell/index.tsx
```diff
@@ -1,34 +1,35 @@
...
+  const [isEditing, setIsEditing] = useState(false);
+  const [value, setValue] = useState<string | undefined>(props.value?.[0]);
 
-  const [isEditing, setIsEditing] = React.useState(false);
-  const [value, setValue] = React.useState<string | undefined>(props.value?.[0]);
+  // Load program options once
+  useEffect(() => {
+    if (!programOptions.length) {
+      dispatch(requestProgramOptions());
```
hughes036, line 27: Looks like this is the most meaningful difference.  Can you explain the big picture change?

src/renderer/services/labkey-client/index.ts
```diff
@@ -306,6 +306,25 @@ export default class LabkeyClient extends HttpCacheClient {
     return response.rows;
   }
 
+  /**
+   * Get all options for program from labkey
+   */
+  public async getProgramOptions(): Promise<AnnotationOption[]> {
+    const filters = ["query.AnnotationId~eq=153"];
```
hughes036, line 313: Looks like this replaces the previous filtering that you did in `index`, and then passes the result to there via an established redux pattern (which i can't follow for the life of me) ?

hughes036 (commented):
This looks right but i have no idea how to review the redux parts.  Is it following a pattern established for barcode / notes?

I believe that a benefit of redux is ease of unit testing.  Do the other cases following this pattern have tests that you should implement?

src/renderer/services/labkey-client/index.ts
tyler-foster: > previously the options came with the template via annotationOptions

Huh, interesting. There are currently only 645 "AnnotationOption" rows -> In other places, do we just filter on those when we need them? Seems a little clunky.

Nitpicky, but I'd probably take out the `153` on this line and replace it with a `PROGRAM_ANNOTATION_ID` constant, just to make the query a bit more obvious.

tyler-foster (changes requested):
Tested it out, works for me. Good job!

Don't you love all the Redux / Redux-Logics boilerplate? Isn't it fun to write all that stuff out four times?

Requesting changes because we should at least add unit tests to `/state/metadata/logics.test.ts` and `/state/metadata/reducer.test.ts`. They should help if we mess with the structure of state at all.

hughes036 (approved):

tyler-foster (approved):
Thanks for adding the unit tests, LGTM

PR #198, opened 2025-07-22
Fix getAnnotations to also attach program annotations

yrkim98:
Program annotations were not being uploaded with the rest of the annotations even when selected.

Since we are enforcing that program is filled out on the frontend (see https://github.com/aics-int/aics-file-upload-app/pull/193) - a db change was made so that `exposeToFileUploadApp` for program annotations is now `false`. That also made it so that it wasnt returned as one of the annotations included in a template.

`getAnnotations()` for the upload only includes annotations that are part of a template, so program was being from uploaded annotations.

`getAnnotations()` now manually checks for a program value, and if one was selected, it is included with the rest of the annotations gathered there.

### DIFF
--- a/src/renderer/state/upload/test/selectors.test.ts
+++ b/src/renderer/state/upload/test/selectors.test.ts
@@ -43,7 +43,7 @@ import {
   getUploadRequests,
 } from "../selectors";
 import { getUploadAsTableRows, getUploadValidationErrors, getAnnotations } from "../selectors";
-import { FileType, MMSAnnotationValueRequest } from "../types";
+import { FileType } from "../types";
 
 
 // utility function to allow us to deeply compare expected and actual output without worrying about order


src/renderer/state/upload/selectors.ts
```diff
@@ -436,6 +437,24 @@ const getAnnotations = (
     },
     [] as MMSAnnotationValueRequest[]
   );
+  const programValue = fileMetadata[AnnotationName.PROGRAM];
+  const programPresent = Array.isArray(programValue)
```
saeliddp, line 441: JS noob question: can `programValue` either be an array or `null`? Or does `isNil` handle some other cases as well?

tyler-foster (changes requested):
I can confirm this works when a program value is provided!

Change request:
* Please add / update a unit test for `/state/upload/selectors.ts`

saeliddp (approved):

tyler-foster (approved):

Issue #199, opened 2025-08-13
Fix mocha tests with 22.18 node changes


yrkim98:
Node 22.18 enables type stripping by default, breaking our mocha tests. (Apparently node will try to run test `.ts` files itself without letting `mocha` compile them, which causes some test fixtures to not load properly.

We'll have to disable this for our test files, and unpin 22.17 which I pinned as a short term fix.

hughes036:
Didnt this pr get merged?

PR #200, opened 2025-08-13
Fix broken build

yrkim98:
node `22.18` introduces [type stripping enabled by default](https://github.com/nodejs/node/releases?utm_source=chatgpt.com) which is breaking our mocha tests.

Discussed with @tyler-foster and for an immediate build fix I am pinning node 22.17 for now. Created a ticket with details for a fix soon https://github.com/aics-int/aics-file-upload-app/issues/199.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ jobs:
       - name: Initialize
         uses: actions/setup-node@v1
         with:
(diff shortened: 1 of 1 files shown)


tyler-foster (approved):
A lack of version pinning seems to bite us in the butt every few months 🙁 . At least the fix was easy

PR #203, opened 2025-09-15
Update FUA to use FSS v4

yrkim98:
Big update to update fua to use FSS v4.

Main thing changed is removing anything to do with chunked uploads, using fssv4's new upload endpoint instead.

Most important files to look at are the following:

`src/renderer/services/file-management-system/index.ts`: contains the bulk of the changes needed to match FUA up with FSS v4. Mainly removing old v3 methods (`register`, `finalize`) and chunked uploads and replacing them with new v4 upload method. Also some changes to `retry` and `cancel` which were simplified in FSS v4.
`src/renderer/services/file-storage-service/index.ts`: same thing

`src/renderer/containers/App/handleUploadJobUpdates.ts`: handles progress updates as upload progresses

`src/main/index.ts`: changed to re-enable react dev tools with a stackoverflow post i found
`package.json`: changed to re-enable react dev tools with a stackoverflow post i found
`package-lock.json`: added `concurrently` and `wait-on` which added a bunch of deps, is this normal for npm?

the rest of the files is updating tests to v4 or removing tests that test for old v3 patterns.

note: I was trying vscode's source control for the first time and thought i was committing small chunks of this but ig I was not and everything was squashed into one commit at the end

### DIFF
--- a/src/renderer/containers/App/test/App.test.ts
+++ b/src/renderer/containers/App/test/App.test.ts
@@ -58,36 +58,4 @@ describe("App", () => {
             });
         });
     });
-    it("dispatches updateUploadProgressInfo when multifile progress is updated", () => {
-        // Arrange
-        const fssJob: JSSJob = {
-            created: new Date(),
-            jobId: "foo123",
-            jobName: "test_file.txt",
-            modified: new Date(),
-            originationHost: "dev-aics-fup-001",
-            service: Service.FILE_STORAGE_SERVICE,
-            updateParent: false,
-            user: "fakeuser",
-            status: JSSJobStatus.WORKING,
-            serviceFields: {
-                multifile: true,
-                fileSize: 100,
-                subfiles: {
-                    'fileid1': 10,
-                    'fileid2': 10,
-                    'fileid3': 15
-                },
-            },
-        };
-        let actionPersisted = undefined;
-        const dispatch = (action: Action)=>{
-            actionPersisted = action;
-        };
-        const expectedAction = updateUploadProgressInfo(fssJob.jobId, { bytesUploaded: 35, totalBytes: fssJob.serviceFields?.fileSize, step: Step.TWO })
-        // Act
-        handleUploadJobUpdates(fssJob, dispatch);
-        // Assert
-        expect(actionPersisted).to.deep.equal(expectedAction);
-    });
 });


src/renderer/services/file-storage-service/index.ts
```diff
@@ -69,9 +50,10 @@ interface FileRecord {
  * This acts as an interface for interacting with the File Storage Service (FSS).
  */
 export default class FileStorageService extends HttpCacheClient {
-  public static readonly ENDPOINT = "fss2/v3.0";
-  private static readonly BASE_FILE_PATH = `${FileStorageService.ENDPOINT}/file`;
-  private static readonly BASE_UPLOAD_PATH = `${FileStorageService.ENDPOINT}/upload`;
+  public static readonly ENDPOINT_V4 = "fss2/v4.0";
+  public static readonly ENDPOINT_V3 = "fss2/v3.0";
```
kmitcham, line 54: should the V3 be deleted from here and the readme?

kmitcham (approved):

src/renderer/services/file-storage-service/index.ts
```diff
@@ -92,32 +74,27 @@ export default class FileStorageService extends HttpCacheClient {
...
-      // Unfortunately FSS expects snake_case
-      // so the conversion must be manual each request
-      file_size: fileSize,
-      local_nas_shortcut: localNasPath !== undefined,
-      local_nas_path: localNasPath,
+      fileName,
+      fileType,
+      source,
+      path,
       multifile: !!isMultifile,
```
saeliddp, line 96: What's with the double negation? Is it a way to handle `isMultifile = null`?

src/renderer/services/file-storage-service/index.ts
hughes036: @saeliddp Do you think we should consider each resource to have its own version system ?  That seems more confusing for clients.

src/renderer/services/file-management-system/index.ts
```diff
@@ -253,117 +149,39 @@ export default class FileManagementSystem {
...
+      this.succeedUpload(uploadId,
+        fuaUpload.serviceFields.result?.[0].fileId || '',
+        fuaUpload.serviceFields.result?.[0].fileName || '',
+        fuaUpload.serviceFields.result?.[0].readPath || '',
       );
+      throw new Error(`Upload cannot be retried if already successful.`);
     }
 
-    // Attempt to resume an ongoing upload if possible before scraping this one entirely
     const { fssUploadId } = fuaUpload.serviceFields;
```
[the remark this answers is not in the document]
hughes036: @saeliddp https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring

src/renderer/services/file-storage-service/index.ts
hughes036: Yeah this looks like a possible bug @yrkim98

hughes036:
@saeliddp We wont have this situation, but from an engineering pov, how someone writing a new query find out what version to use for each resource?  My sense is that we should just roll fwd the whole API version (even resources that have not changed), and document what our current version is on the FSS readme.

Thoughts?

src/renderer/services/file-storage-service/index.ts
```diff
@@ -11,12 +11,12 @@ import { HttpClient } from "../types";
...
-    preUploadMd5?: number;
-    postUploadMd5?: number;
     fileSize?: number;
-    currentFileSize?: number;
+    copyToFmsCacheProgress?: number; // hybrid only
+    checksumProgress?: number;       // cloud + hybrid
+    s3UploadProgress?: number;       // final stage
     multifile?: boolean;
-    subfiles?: { [fileId: string]: number }; // mapping of subfile IDs to their respective # of bytes uploaded
+    subfiles?: { [fileId: string]: number };
```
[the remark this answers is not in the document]
saeliddp: Is it safe to remove `subfiles?: { [fileId: string]: number };`, not just the comment?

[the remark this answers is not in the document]
saeliddp: ^ also a GH diff issue I think

Issue #131, opened 2025-01-29
Dev: Pre-commit hooks not working


tyler-foster:
I'm fairly certain I broke the Husky config way back in June '24 when I made the big update to how the FUA build process works. I made [some progress on fixing this](https://github.com/aics-int/aics-file-upload-app/pull/90) but never finished the work. Should reinvestigate again so we can maintain basic standards like linting / formatting.

PR #214, opened 2025-10-15
Applying prettier formatting.

hughes036:
This pr simply adds an npm script for running prettier, and applies the result of running it.


hughes036:
> Side note, this jogged my memory, and I realized I made this issue months ago: #131

Handled this in the followup pr.

Issue #226, opened 2025-12-05
BUG: closing FUA during uploads


saeliddp:
From [this sw support thread](https://allencellscience.slack.com/archives/C021CEY96LD/p1764959403675449):

Rennie observed the following:
- started a few uploads
- during the uploads, turned off computer
- upon reopening FUA, see 0% uploaded as the status
- attempt to cancel, but nothing happened

Affected files:
```
3500008143_20251111_20X_Timelapse-01(74).czi
3500008143_20251111_20X_Timelapse-01(75).czi
3500008143_20251111_20X_Timelapse-01(72).czi
3500008143_20251111_20X_Timelapse-01(73).czi
```

From the FSS logs, we see these messages, corresponding to each file that Rennie sees at 0%:
```
 {"@timestamp":"2025-12-05T18:42:37.144+00:00","@version":"1","message":"Returning 409 CONFLICT to client due to error:\n Upload's current status conflicts with requested action (Attempted to retry upload 335f6823a86e482c9284d175f9548d84 with status COMPLETE)","logger_name":"org.alleninstitute.aics.fss.controller.UploadController","thread_name":"https-jsse-nio-9977-exec-11","level":"ERROR","level_value":40000,"traceId":"ff2f6d30b1bb1b57","spanId":"ff2f6d30b1bb1b57"}
```

The JSS jobs, and statuses in FSS DB all appear to be normal.

**Theory**:
It seems like FUA either:
- has stale JSS data about the uploads in question
- fails to update the UI to reflect new data on a restart

Additionally, it seems like we have automatic retries (I thought we got rid of that behavior?)

PR #227, opened 2025-12-12
Disable automatic retries on app restart, update abandoned job statuses on restart

yrkim98:
### Summary
This PR addresses [this bug reported by](https://github.com/aics-int/aics-file-upload-app/issues/226) @saeliddp.

### Problem
Before, when the user quits FUA during an upload creating an abandoned job:

- The job would finish while the app is closed and FSS would mark as complete
- The app would retry all abandoned jobs on restart
- FSS would return 409 CONFLICT
- UI would never update to show the completion

### Solution

- Disabled retries on app restart for any abandoned jobs- the user should retry or cancel any failed ones manually like we tell them to do when they quit the app during an upload.
image

- Check abandoned jobs status on app restart and mark complete if they finished while the app was closed

### Changes

- `src/renderer/services/file-management-system/index.ts`: added a function `isAbandonedJobComplete()` which helps sync upload statuses for abandoned jobs on app restart. Checks jss, then fss, and marks `complete()` if job is finished
- `src/renderer/state/job/logics.ts`: No longer calls `retry()`, instead calls `isAbandonedJobComplete()`
- tests updated to no longer test for retry and instead see if file sync happens in `src/renderer/state/job/logics.ts`

### Testing
Confirmed working- I started a larger upload, quit the app while upload was in progress, restarted app after upload was complete on fms, and verified the upload refreshed to complete when the app was reopened.

### DIFF
--- a/src/renderer/services/file-management-system/index.ts
+++ b/src/renderer/services/file-management-system/index.ts
@@ -215,7 +215,12 @@ export default class FileManagementSystem {
 
     const fssStatus = await this.fss.getStatus(fssUploadId);
 
-    if (fssStatus.status === UploadStatus.COMPLETE && fssStatus.fileId) {
+    if (fssStatus.status === UploadStatus.COMPLETE) {
+      if (!fssStatus.fileId) {
+        throw new Error(
+          `FSS upload ${fssUploadId} is COMPLETE but missing fileId?`
+        );
+      }
       // fss reporting complete
       await this.complete(fuaUpload, fssStatus.fileId);
       return true;


saeliddp:
> like we tell them to do when they quit the app

The message in your screenshot looks out of date. I don't think we need any warning now, since the upload happens completely async and FUA just listens for JSS updates. What do you think?

I'm also wondering if 'abandoned' is a status that the FUA JSS job tracks? If so, that also seems out of date but maybe that would be part of the larger FUA/JSS refactor effort that we've discussed a bit in standup.

src/renderer/services/file-management-system/index.ts
```diff
@@ -193,6 +193,38 @@ export default class FileManagementSystem {
...
+
+    const fssStatus = await this.fss.getStatus(fssUploadId);
+
+    if (fssStatus.status === UploadStatus.COMPLETE && fssStatus.fileId) {
+      // fss reporting complete
+      await this.complete(fuaUpload, fssStatus.fileId);
+      return true;
+    }
+
+    // upload is still in progress
```
saeliddp: How does a still-in-progress upload get its progress indicator updated?

hughes036:
> Confirmed working- I started a larger upload, quit the app while upload was in progress, restarted app after upload was complete on fms, and verified the upload refreshed to complete when the app was reopened.

Did you also confirm that this test reproduces the bug behavior before the change is applied?

src/renderer/services/file-management-system/index.ts
hughes036: I think a subscription to JSS would need to be made.

hughes036 (approved):
There are a few more improvements (commented by @saeliddp ) that I agree should be made.  But lets balance that with getting the bugfix out asap.

Can you test what happens when an in progress upload is found by `isAbandonedJobComplete` ?  Do we have a mechanism for then subscribing and showing progress?

PR #239, opened 2026-02-03
Fix selection dropdowns

yrkim98:
This PR addresses a bug with https://github.com/orgs/aics-int/projects/30/views/1?pane=issue&itemId=71300273&issue=aics-int%7Caics-file-upload-app%7C101 and https://github.com/orgs/aics-int/projects/30/views/1?pane=issue&itemId=70688419&issue=aics-int%7Caics-file-upload-app%7C95

### Issue 1: allow arrow key use for dropdown input boxes
Currently, if the users use keyboard arrow down/up/enter inputs to make a selection in the dropdown, the value does not get populated into the table's cell.

**Problem**
It looks like what was happening was that when the user used arrow keys to make a selection, the `DropdownEditor`'s `onInputKeyDown` handler was firing first and causing `handleCommit()` to commit an empty value to the cell. Then the components internal enter handling fires afterwards, and also commits an empty value to the cell.

**Solution**
we pretty much have a redundant `onInputKeyDown` handler that is interfering with the built in native handler that comes with the `Select` component. Removing fixes the issue. I dont see `onInputKeyDown`  being used for anything else so removing should have no side effects

### Issue 2: PlateBarcodeCell dropdown issue
Currently, if the users select a Plate Barcode and click it again to change it then click off, the field updates to `[object Object]`. See here for example https://github.com/orgs/aics-int/projects/30/views/1?pane=issue&itemId=70688419&issue=aics-int%7Caics-file-upload-app%7C95

**Problem**
when the user clicks off the input field, the `onBlur` event was passing a `FocusEvent` object to `onCommit`. Since `onCommit` uses this value as the barcode value, it was receiving the event object and displaying that as the barcode value.

Solution
if we wrap `onCommit`, the event will never get passed to `onCommit` and `onCommit` will use the default `plateBarcode` value every time like it's supposed to.

### DIFF
--- a/src/renderer/containers/LookupSearch/index.tsx
+++ b/src/renderer/containers/LookupSearch/index.tsx
@@ -36,7 +36,6 @@ interface OwnProps {
   lookupAnnotationName: keyof MetadataStateBranch;
   lookupTable?: string;
   onBlur?: () => void;
-  onInputKeyDown?: (e: React.KeyboardEvent) => void;
   placeholder?: string;
 }
 
@@ -136,7 +135,6 @@ class LookupSearch extends React.Component<Props, { searchValue?: string }> {
           mode={mode}
           dropdownRender={this.props.dropdownRender}
           notFoundContent={notFoundContent}
-          onInputKeyDown={this.props.onInputKeyDown}
           onBlur={onBlur}
           // Unable to type this variable most explicitly due to "type narrowing in discriminated unions"
           // typing as any for now though actual value will either be a string or string[]


hughes036 (approved):

src/renderer/containers/Table/CustomCells/PlateBarcodeCell/index.tsx
```diff
@@ -130,7 +130,7 @@ export default function PlateBarcodeCell(
         retrieveOptionsOverride={(i?: string) =>
           i && dispatch(requestBarcodeSearchResults(i))
         }
-        onBlur={onCommit}
+        onBlur={() => onCommit()}
```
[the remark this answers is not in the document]
tyler-foster: > There's one case i can see where we actually do need to pass args

Which one is that?

~~In this specific case, was the problem that a blur `Event` object was being passed to `onCommit()`?~~ Read the PR description lol

tyler-foster (changes requested):
The `FocusEvent` thing is probably good, just asked one question in the comment chain w/ Chris.

Re: the `<DropdownEditor>` component, I left another comment about how the issue that cropped up there might be present elsewhere. Could you check that for me? Otherwise I believe it LGTM.

src/renderer/containers/Table/CustomCells/PlateBarcodeCell/index.tsx
```diff
@@ -130,8 +129,7 @@ export default function PlateBarcodeCell(
         retrieveOptionsOverride={(i?: string) =>
           i && dispatch(requestBarcodeSearchResults(i))
         }
-        onBlur={onCommit}
-        onInputKeyDown={createEnterKeyHandler(onCommit)}
```
tyler-foster: I believe there's a `<Select>` sub component to this that would also need the `onInputKeyDown` removed, similar to the one you updated in `DropdownEditor`

tyler-foster: Oh, and if `<LookupSearch>` is a custom component then you might want to also update its `propTypes` at the bottom.

PR #246, opened 2026-03-03
Use most informative fss error message

yrkim98:
Currently we are using the standard `axios` `error.message` to show the user what went wrong during an upload if it fails. This results in an uninformative error message such as "Request failed with status code 400".

The actual response from fss with the informative error message is in  `error.response.data.message`.

**Changes**
src/renderer/services/file-management-system/index.ts: When fss call fails, error saved to `serviceFields` uses informative `error.response.data.message` if available, otherwise falling back to less informative `error?.message` or `Upload failed` (like before). This will make it so that the more informative message shows up in the upload table for this upload job.

src/renderer/state/upload/logics.ts: `uploadFailed()` and `initiateUploadFailed()` are dispatched with the more informative `error.response.data.message` if available, otherwise falling back to less informative `error?.message`. This makes it so the error message popup shows the more informative message

This results in an error message as follows:
Screenshot 2026-03-03 at 12 12 33 PM

### DIFF
--- a/src/renderer/state/upload/test/logics.test.ts
+++ b/src/renderer/state/upload/test/logics.test.ts
@@ -296,6 +296,9 @@ describe("Upload logics", () => {
         undefined,
         uploadLogics
       );
+      const fileNames = getUploadFileNames(
+        nonEmptyStateForInitiatingUpload
+      ).join(", ");
       const apiMessage = "Path does not exist: /aled/eadf/test";
       const axiosError = new Error(
         "Request failed with status code 400"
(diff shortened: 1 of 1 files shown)


src/renderer/services/file-management-system/index.ts
```diff
@@ -349,7 +349,8 @@ export default class FileManagementSystem {
         status: JSSJobStatus.FAILED,
         serviceFields: {
           ...upload.serviceFields,
-          error: error?.message ?? "Upload failed",
+          error:
+            error?.response?.data?.message || error?.message || "Upload failed",
```
saeliddp: Does this mean that the UI will show `Upload <someuploadname> failed: Upload failed` in the case where we don't have either error message?

And is it even possible to have `error.message` be undefined when we're executing this code?

If it is possible, I suggest changing the last resort string to "Unknown cause"; if it's not possible, seems like we can just get rid of it altogether.

src/renderer/state/upload/logics.ts
```diff
@@ -769,7 +775,9 @@ const uploadWithoutMetadataLogic = createLogic({
       } catch (error) {
         dispatch(
           uploadFailed(
-            `Something went wrong while uploading your files. Details: ${error?.message}`,
+            `Something went wrong while uploading your files. Details: ${
+              error?.response?.data?.message || error?.message
```
saeliddp: I'm unfamiliar with react/ts details, but I see the same/similar logic being repeated multiple times here. Any way to avoid that?

yrkim98: [This](https://github.com/aics-int/aics-file-upload-app/pull/246/commits/643ea3e7536740d39a65b2056392663b3b266102) should tidy things up. :)

hughes036 (approved):

PR #259, opened 2026-03-09
Bugfix: some template items not being autofilled

yrkim98:
**Bug Description**
Some files were not being autofilled with known metadata in the metadata template. I saw that template auto-application was racing with metadata fetching, sometimes causing some files to not have metadata autofilled.

**Fix**
I apply the template on the "Continue to Metadata" button click- since the button is already disabled until all metadata is loaded, this makes sure that a complete autofill of all files is possible.

This should have been done in https://github.com/aics-int/aics-file-upload-app/pull/243 and is therefore an extension of that pr, see for related changes.

https://github.com/user-attachments/assets/c882bf2c-bd9d-4fc6-b7ab-f8814ca419fd

### DIFF
--- a/src/renderer/state/upload/logics.ts
+++ b/src/renderer/state/upload/logics.ts
@@ -38,6 +38,7 @@ import {
   getSelectedUploads,
   getSelectedUser,
 } from "../selection/selectors";
+import { getTemplateId } from "../setting/selectors";
 import { ensureDraftGetsSaved, getApplyTemplateInfo } from "../stateHelpers";
 import { setAppliedTemplate } from "../template/actions";
 import { getAppliedTemplate } from "../template/selectors";
@@ -56,6 +57,7 @@ import { batchActions } from "../util";
 
 import {
   addUploadFiles,
+  applyTemplate,
   cancelUploadFailed,
   cancelUploadSucceeded,
   editFileMetadataFailed,
--- a/src/renderer/state/upload/logics.ts
+++ b/src/renderer/state/upload/logics.ts
@@ -155,10 +155,17 @@ const applyTemplateLogic = createLogic({
 
 const addUploadFilesLogic = createLogic({
   process: (
-    _deps: ReduxLogicProcessDependencies,
-    _dispatch: ReduxLogicNextCb,
+    { getState }: ReduxLogicProcessDependencies,
+    dispatch: ReduxLogicNextCb,
     done: ReduxLogicDoneCb
   ) => {
+    const selectedTemplate = getAppliedTemplate(getState())?.templateId;
+    const savedTemplate = getTemplateId(getState());
+    if (selectedTemplate) {
+      dispatch(applyTemplate(selectedTemplate));
+    } else if (savedTemplate) {
+      dispatch(applyTemplate(savedTemplate));
+    }
     done();
   },
   type: ADD_UPLOAD_FILES,
--- a/src/renderer/state/upload/logics.ts
+++ b/src/renderer/state/upload/logics.ts
@@ -38,7 +38,6 @@ import {
   getSelectedUploads,
   getSelectedUser,
 } from "../selection/selectors";
-import { getTemplateId } from "../setting/selectors";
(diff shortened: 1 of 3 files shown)


hughes036:
Im pretty confused about why these 2 things would be in a race.  Template-auto-application is not async - why would it be called / triggered before the metadata is ready?  Does this seem more complicated than it should be, maybe because of all the different logics in play?  Id be interested in a walk through tomorrow.

yrkim98:
> Im pretty confused about why these 2 things would be in a race. Template-auto-application is not async - why would it be called / triggered before the metadata is ready? Does this seem more complicated than it should be, maybe because of all the different logics in play? Id be interested in a walk through tomorrow.

happy to jump on a call to talk about this, it is pretty confusing

BrianWhitneyAI (approved):

yrkim98:
Ok simplified the logics, and verified everything still works as expected.

I removed the logics that applies a template after each file addition (which is unnecessary, and creates that race) and instead have one applyTemplate call on the "Continue to Metadata" button press when the user is done adding files, and we know we have all metadata back from mxs

src/renderer/state/upload/logics.ts
```diff
@@ -155,24 +151,6 @@ const applyTemplateLogic = createLogic({
   type: APPLY_TEMPLATE,
 });
 
-const addUploadFilesLogic = createLogic({
-  process: (
-    { getState }: ReduxLogicProcessDependencies,
-    dispatch: ReduxLogicNextCb,
-    done: ReduxLogicDoneCb
-  ) => {
-    const selectedTemplate = getAppliedTemplate(getState())?.templateId;
```
hughes036, line 164: Im confused about the meaning of selectedTemplate and savedTemplate.  Doesnt the UX only allow 1 template at a time to be in scope?

src/renderer/state/upload/logics.ts
```diff
@@ -155,24 +151,6 @@ const applyTemplateLogic = createLogic({
...
-    const savedTemplate = getTemplateId(getState());
-    if (selectedTemplate) {
-      dispatch(applyTemplate(selectedTemplate));
-    } else if (savedTemplate) {
-      dispatch(applyTemplate(savedTemplate));
-    }
-    done();
-  },
-  type: ADD_UPLOAD_FILES,
```
hughes036, line 173: > I apply the template on the "Continue to Metadata" button click- since the button is already disabled until all metadata is loaded, this makes sure that a complete autofill of all files is possible.

is `ADD_UPLOAD_FILES` what fires on "Continue to Metadata" button click?

src/renderer/state/upload/test/logics.test.ts
```diff
@@ -888,55 +888,6 @@ describe("Upload logics", () => {
     });
   });
 
-  describe("addUploadFilesLogic", () => {
-    it("applies selected templated over saved template", async () => {
```
hughes036, line 892: Is this when the user changes the template on an already COMPLETE upload?

src/renderer/state/upload/test/logics.test.ts
```diff
@@ -888,55 +888,6 @@ describe("Upload logics", () => {
...
-      // arrange
-      const templateId = 17;
-      const badTemplateId = 4;
```
hughes036, line 895: why is it bad?

src/renderer/state/upload/test/logics.test.ts
```diff
@@ -888,55 +888,6 @@ describe("Upload logics", () => {
...
-        },
-      });
-
-      // act
-      store.dispatch(addUploadFiles([]));
-      await logicMiddleware.whenComplete();
-
-      // assert
-      expect(actions.includesMatch(applyTemplate(badTemplateId))).to.be.false;
-      expect(actions.includesMatch(applyTemplate(templateId))).to.be.true;
```
hughes036, line 917: Is this the same thing that the test below is covering?

hughes036 (commented):
I could use another in person walk through

hughes036 (approved):
reviewed in person

PR #265, opened 2026-03-17
Change cell error flash color

yrkim98:
Travis requested that the flashing red color be changed for the cell error.

previously we were using the default error color, now changed to #FEE0DE.

(It looks like we use the default red color in a couple other css files, but maybe we should change all error colors to #FEE0DE? For now just doing it in this one place.)

The new color looks like this:

https://github.com/user-attachments/assets/2591d285-b125-4ea2-bbad-f7637f8e6417

### DIFF
--- a/src/renderer/containers/Table/DefaultCells/DisplayCell/styles.pcss
+++ b/src/renderer/containers/Table/DefaultCells/DisplayCell/styles.pcss
@@ -69,7 +69,7 @@
 }
 
 @keyframes flash-error {
-    0% { background-color: var(--error); }
+    0% { background-color: #FEE0DE; }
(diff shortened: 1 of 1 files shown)


TravisKroeker (approved):
If we change the default error color, it should be changed to #F92D20. The lighter red is you implemented is a variant of the same red color ramp as #F92D20, but is too light for text or most other warning applications. The reason I changed it in this PR was that the dark red was too alarming- especially in conjunction with the popover.

hughes036 (approved):
I think the next time we meet, we should start building a workflow where @TravisKroeker can make changes like this directly and pr them.  In the age of AI agents we can have a much tighter feedback loop.

PR #269, opened 2026-03-24
auto select well from extracted metadata

yrkim98:
**Summary**                                                                                                                                                                   
@antoineborensztejn requested automatically filling well metadata after the user enters a barcode.
 I made a change so that when a plate barcode is set and MXS has extracted Row and Column metadata for a file,  it will automatically select the matching well

https://github.com/user-attachments/assets/f3c155cf-a4b5-4c89-bf28-11c535040e4d

 
**Changes**
src/renderer/state/upload/logics.ts: added code to autofill well inside the `updateUploadLogic()`, which fires whenever the template is edited. Specifically, the code i added runs when plate barcode is entered and validated.

I'm piggybacking on existing code that sets plate/well info for the well selector UI. Previously, this was just set so that once the user entered a barcode, they could click on the well cell and select a well. With this addition, if the image has valid `Row` and `Col` in its metadata, it will automatically select the corresponding well.

src/renderer/state/upload/test/logics.test.ts: unit tests for this change. Mainly testing the well autoselection, making sure to only autoselect if row/col metadata exists, and making sure to select only valid wells.

**Test**
- Tested with images from VAST with known wells: confirmed correct, well auto-selected                                                      
- End to end test confirmed MMS received the auto-selected well and updated in BFF
- Would like to get a pre-release version to Hannah Thorpe to make sure this works for her data before releasing

**Note on well indexing**
using well A1 as an example:

- In the image metadata this will be row = 1, col = 1
- FUA converts it to a 0 index
- We select 0,0 in the plate layout -> well A1 -> update annotations that will be sent to MMS with this well ID
- This is in line with what MXS does when extracting well IDs in its PATCH endpoint

**Note on not greying out autofilled wells**

- FUA currently greys out and disables all autofilled cells
- I chose not to do this for wells bc I saw some instances during testing such as [this image](https://bff.allencell.org/app?c=file_name%3A0.4%2CKind%3A0.2%2CType%3A0.25%2Cfile_size%3A0.15&filter=%7B%22name%22%3A%22file_name%22%2C%22value%22%3A%223500007829_20251031_20X_Timelapse-01.czi%22%2C%22type%22%3A%22default%22%7D&source=%7B%22name%22%3A%22AICS+FMS%22%7D&sort=%7B%22annotationName%22%3A%22uploaded%22%2C%22order%22%3A%22DESC%22%7D) where there is only one row, col pair in the image metadata yet the upload user selected multiple wells.


hughes036:
> This is in line with what MXS does when extracting well IDs in its PATCH endpoint

Does MXS do this on PATCH only (not PUT)?

We haver a ticket for making PUT behave more like PATH, could WellId piggyback?

src/renderer/state/upload/logics.ts
```diff
@@ -368,7 +391,7 @@ const updateUploadLogic = createLogic({
     // If a plate barcode is being updated check for imaging sessions
     const plateBarcode = upload[AnnotationName.PLATE_BARCODE]?.[0];
     if (plateBarcode) {
-      const plateBarcodeToPlates = getPlateBarcodeToPlates(deps.getState());
+      let plateBarcodeToPlates = getPlateBarcodeToPlates(deps.getState());
```
hughes036: Style: Rather than reassign `plateBarcodeToPlates`, I would just create a new variable below.

src/renderer/state/upload/logics.ts
```diff
@@ -398,12 +421,34 @@ const updateUploadLogic = createLogic({
...
-          })
+        plateBarcodeToPlates = {
+          ...plateBarcodeToPlates,
+          [plateBarcode]: imagingSessionsWithPlateInfo,
+        };
+        dispatch(setPlateBarcodeToPlates(plateBarcodeToPlates));
+      }
+
+      // autoselect well if row and col data available from mxs
+      const fileKey = deps.action.payload.key;
```
hughes036, line 433: Logics noob question - execution `dispatch` is async, and execution continues below, right?

Having a single logic do so much, including dispatch multiple events conditionally, seems like a great way to create complex side effects that interact in ways that are difficult to replicate / reason about.

hughes036: Does `dispatch(setPlateBarcodeToPlates(plateBarcodeToPlates));` just set the barcode in the template table?

And everything below here is just about setting the well in the template?

If so, could you just update state & emit 1 event?  And break some if this login into local methods for readability?

src/renderer/state/upload/logics.ts
```diff
@@ -398,12 +421,34 @@ const updateUploadLogic = createLogic({
...
+      const mxsData = deps.getState().metadataExtraction[fileKey]?.metadata;
+      const rowValue = mxsData?.["Row"]?.value;
+      const colValue = mxsData?.["Column"]?.value;
+
+      if (rowValue !== undefined && colValue !== undefined) {
+        // row and col are 1-indexed in metadata, convert to 0-indexed
+        const row = Number(rowValue) - 1;
+        const col = Number(colValue) - 1;
+        const plates = plateBarcodeToPlates[plateBarcode];
+        const well = plates?.[0]?.wells.find(
```
BrianWhitneyAI: Should there be some error case around if multiple plates are returned? I remember this being an issue at one point

BrianWhitneyAI (approved):

src/renderer/state/upload/logics.ts
kmitcham: If multiple plates are found, failing the upload makes a certain amount of sense.  I don't think there are supposed to be repeated barcodes.

kmitcham (approved):

hughes036:
> Note on not greying out autofilled wells

> 
> FUA currently greys out and disables all autofilled cells
> I chose not to do this for wells bc I saw some instances during testing such as [this image](https://bff.allencell.org/app?c=file_name%3A0.4%2CKind%3A0.2%2CType%3A0.25%2Cfile_size%3A0.15&filter=%7B%22name%22%3A%22file_name%22%2C%22value%22%3A%223500007829_20251031_20X_Timelapse-01.czi%22%2C%22type%22%3A%22default%22%7D&source=%7B%22name%22%3A%22AICS+FMS%22%7D&sort=%7B%22annotationName%22%3A%22uploaded%22%2C%22order%22%3A%22DESC%22%7D) where there is only one row, col pair in the image metadata yet the upload user selected multiple wells.

Looks like you still need to implement the new decision about 'graying out' well auto-selections; LMK when u are ready for another review.

src/renderer/state/upload/logics.ts
yrkim98: https://github.com/aics-int/aics-file-upload-app/pull/269/changes/3964be7b9a2e5313edaf6b99c6cef86491dde25a

src/renderer/state/upload/logics.ts
yrkim98: `redux-logic`'s `disptch` is async + you are right in that it is confusing but in this case I don't see it creating any side effects, the two dispatches i added dont interact at all
`dispatch(setPlateBarcodeToPlates())` updates state.metadata  for well lookup data, which later renders the plate UI
`dispatch(updateUpload())` updates state.upload for the well autofill
and the well autofill reads from a local variable, instead of state.metadata, so it wont matter if that dispatch hasn't hit yet.

It would be difficult to couple this into one event since they update different state. We'd have to merge the two separate states and have its reducer handle that event, and IMO it makes sense to have these separated

but happy to break this method down into helpers for readability

src/renderer/state/upload/logics.ts
yrkim98: this array is `PlateImagingSession[]`, not barcode (so if multiple imaging sessions have the same plate). I update the code to match existing code we have in WellCell on how to handle this case, rather than blindly selecting the first one

yrkim98: the multple barcodes issue seems like a data issue, not a fua issue, so leaving it for now.

src/renderer/containers/Table/CustomCells/WellCell/index.tsx
```diff
@@ -156,7 +158,9 @@ export default function WellCell(props: CellProps<FileModel>) {
         value={wellLabels.sort()}
         disabled={!wells?.[0].length}
         onTabExit={() => setIsEditing(false)}
-        onStartEditing={() => wells?.[0].length && setIsEditing(true)}
+        onStartEditing={() =>
+          !isAutofilled && wells?.[0].length && setIsEditing(true)
```
hughes036, line 162: Does `setIsEditing(true)` render the well picker?

src/renderer/state/upload/reducer.ts
```diff
@@ -83,11 +83,24 @@ const actionToConfigMap: TypeToDescriptionMap<UploadStateBranch> = {
         };
       }
 
+      const currentUpload = state[action.payload.key];
+      const mergedAutofilledFields = upload.autofilledFields
+        ? [
+            ...new Set([
+              ...(currentUpload.autofilledFields || []),
```
hughes036, line 90: This looks like it handles the case when autofill is happening, and the current upload already has autofilled fields.  Is that right?  

What state / workflow leads to this?  

This looks like a good change to comment.

src/renderer/state/upload/reducer.ts
```diff
@@ -83,11 +83,24 @@ const actionToConfigMap: TypeToDescriptionMap<UploadStateBranch> = {
...
+              ...upload.autofilledFields,
+            ]),
+          ]
+        : currentUpload.autofilledFields;
+
       return {
         ...state,
         [action.payload.key]: {
-          ...state[action.payload.key],
+          ...currentUpload,
```
hughes036, line 99: How will upload and currentUpload be different?

src/renderer/state/upload/logics.ts
```diff
@@ -438,14 +439,18 @@ const updateUploadLogic = createLogic({
...
+        const imagingSessionName = upload[AnnotationName.IMAGING_SESSION]?.[0];
+        const plate = plates?.find((p) =>
+          imagingSessionName ? p.name === imagingSessionName : !p.name
         );
+        const well = plate?.wells.find((w) => w.row === row && w.col === col);
         if (well) {
           dispatch(
             updateUpload(fileKey, {
               [AnnotationName.WELL]: [well.wellId],
+              autofilledFields: [AnnotationName.WELL],
```
hughes036, line 453: The effect is, all autofilled fields (including Well) are grayed out.  

When this condition is skipped, the well picker will behave normally.

?

src/renderer/state/upload/test/logics.test.ts
```diff
@@ -828,6 +828,207 @@ describe("Upload logics", () => {
...
+        ],
+      });
+
+      // Act
+      store.dispatch(
+        updateUpload(uploadRowKey, {
+          [AnnotationName.PLATE_BARCODE]: [plateBarcode],
+        })
+      );
+      await logicMiddleware.whenComplete();
```
hughes036, line 898: Nice job on this test.  Very readable.

hughes036 (approved):

PR #276, opened 2026-04-08
Upload extracted metadata from clients

yrkim98:
**Description**
this is the start of our design change to move away from automatic extractions on the backend and start to upload extracted metadata from the clients.

the metadata is still extracted by MXS, but the client calls MXS and we use the client to pass the data forwards to MMS.

**Changes**
- src/renderer/state/metadataExtraction/logics.ts: Identifies which annotations in template are `Duration` type, and converts metadata recieved from MXS (which is in ms) to make sure UI renders it correctly
- src/renderer/state/upload/selectors.ts: Removed exclusion of autofilled metadata, which we now include in the metadata sent to MMS. Merges all MXS extracted metadata with user entered
- src/renderer/state/upload/test/selectors.test.ts: test changes
- src/renderer/state/upload/reducer.ts: Comment change (this will change later again)

**Validation**
Tested with array of czi and nd2 images on staging- uploads show all metadata, user entered, and extracted, on FMS

### DIFF
--- a/src/renderer/state/metadataExtraction/logics.ts
+++ b/src/renderer/state/metadataExtraction/logics.ts
@@ -1,5 +1,8 @@
 import { createLogic } from "redux-logic";
 
+import { DAY_AS_MS, HOUR_AS_MS, MINUTE_AS_MS } from "../../constants";
+import { Duration } from "../../types";
+import { getDurationAnnotationTypeId } from "../metadata/selectors";
 import {
   fetchMetadataSucceeded,
   fetchMetadataFailed,
@@ -18,6 +21,18 @@ import {
 import { autofillFromMXS } from "../upload/actions";
 import { ADD_UPLOAD_FILES } from "../upload/constants";
 import { AddUploadFilesAction } from "../upload/types";
+
+function msToDuration(ms: number): Duration {
+  let remaining = ms;
+  const days = Math.floor(remaining / DAY_AS_MS);
+  remaining -= days * DAY_AS_MS;
+  const hours = Math.floor(remaining / HOUR_AS_MS);
+  remaining -= hours * HOUR_AS_MS;
+  const minutes = Math.floor(remaining / MINUTE_AS_MS);
+  remaining -= minutes * MINUTE_AS_MS;
+  const seconds = Math.floor(remaining / 1000);
+  return { days, hours, minutes, seconds };
+}
 
 // fetches MXS data and stores in metadataExtraction state
 const fetchMetadataLogic = createLogic({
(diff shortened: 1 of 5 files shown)


kmitcham (approved):

src/renderer/state/metadataExtraction/logics.ts
```diff
@@ -19,6 +22,18 @@ import { autofillFromMXS } from "../upload/actions";
 import { ADD_UPLOAD_FILES } from "../upload/constants";
 import { AddUploadFilesAction } from "../upload/types";
 
+function msToDuration(ms: number): Duration {
```
hughes036: Request: unit test this.

hughes036: nit: refactor repeated operation to be DRY.

src/renderer/state/metadataExtraction/logics.ts
```diff
@@ -72,15 +87,34 @@ const autofillOnTemplateAppliedLogic = createLogic({
...
+      (template?.annotations || [])
+        .filter((a) => a.annotationTypeId === durationTypeId)
+        .map((a) => a.name)
+    );
+
     for (const filePath of Object.keys(uploads)) {
       const cachedMetadata = state.metadataExtraction[filePath]?.metadata;
       if (cachedMetadata) {
-        dispatch(autofillFromMXS(filePath, cachedMetadata));
+        const convertedMetadata = Object.fromEntries(
```
hughes036: request: refactor
(this is a pretty confusing statement)

src/renderer/state/upload/selectors.ts
```diff
@@ -469,45 +464,89 @@ export const getAnnotations = (
...
-        ...(fileMetadata[AnnotationName.WELL]?.length && {
-          wellIds: fileMetadata[AnnotationName.WELL],
-        }),
-      },
-    }));
+    const templateAnnotationNames = new Set(
+      template.annotations.map((a) => a.name)
+    );
+
+    return Object.entries(uploads).map(([filePath, fileMetadata]) => {
```
hughes036, line 494: request: refactor

src/renderer/state/metadataExtraction/logics.ts
yrkim98: https://github.com/aics-int/aics-file-upload-app/pull/276/commits/e9483641245c957b7052af54d259520a21cb63f7 added

src/renderer/state/metadataExtraction/logics.ts
yrkim98: https://github.com/aics-int/aics-file-upload-app/pull/276/commits/7d0c1584ae50b905771e5b3454455af37b8cfe67

 this should make it clearer, avoided using map with the logic inside of it and just used a loop, also added comments

src/renderer/state/upload/selectors.ts
yrkim98: I refactored the way we merge metadata, so its done a bit nicer: https://github.com/aics-int/aics-file-upload-app/pull/276/commits/72d339497d18932eeb357ff182d599748cd79624

basically getting extracted metadata, merging in the the user entered metadata if not already present, and using that as the annotations list that gets sent to MMS. Left some comments to clear it up

re the `Object.entries` statement used and the rest of these changes, this was already existing code (see https://github.com/aics-int/aics-file-upload-app/pull/276/changes#diff-84062e0be9e2e33a55a2b5c26fa52316aa958e42eeea3d70925803409ec6cb3dL483)
so i am leaving for now

src/renderer/state/upload/selectors.ts
```diff
@@ -469,45 +464,75 @@ export const getAnnotations = (
   return annotations;
 };
 
+export const getExtractedAnnotations = (
```
hughes036, line 467: naming nit: this seems like a transformation function than a getter.  `normalizeExtractedAnnotations` ?

hughes036: Is this really just a private method?  I only see it called from within this class.  Can you just declare it const and not export?

hughes036 (approved):
Much simpler and easier to read!
Nice work.

src/renderer/state/upload/selectors.ts
yrkim98: its called in the test for this method, so need this

yrkim98: naming change is valid, ill make that change

PR #280, opened 2026-05-01
Bugfix/upload mass editor bugs

yrkim98:
###  Description                                                                                                                                                     
This is a set of fixes to make the annotation mass editor behave consistently with annotating individual rows in the upload template.                                                        
                                                                                                                                                                  
Previously, entering a plate barcode in the mass editor didn't trigger well autofill per file.

additionally, using the mass edit feature or dragging to copy values down a column could overwrite cells that had been autofilled by MXS.                                        
                                                                                                                                                                  
### Changes   
**src/renderer/state/selection/logics.ts**                                                                                                                         
Main changes here is to protect against editing already autofilled rows when drag and dropping or using mass edit UI to change multiple rows in a column.

- disable mass edit if a selected row has already been autofilled                     
- drag to copy skips autofilled rows                

**src/renderer/state/upload/logics.ts**
changes here are for the well autofill

- fix plate barcode not triggering well autofill on re-selection, now autofills the well 
                                                                                                                                                                                                                                                                                          
### Validation                                                                                                                                                      
                                                                                                                                                                  

- tested with czi files that have row/col metadata from MXS. barcodes entered in both single-row and mass edit correctly grey and autofill the well column.      

- dragging over autofilled rows leaves them untouched.

### DIFF
--- a/src/renderer/state/selection/logics.ts
+++ b/src/renderer/state/selection/logics.ts
@@ -177,7 +177,7 @@ const stopCellDragLogic = createLogic({
       const columnId = cellAtDragStart.columnId;
       const rowIds = rows
         .map((row: UploadRowTableId) => row.id)
-        .filter((id) => !upload[id]?.autofilledFields?.includes(columnId));
+        .filter((id: any) => !upload[id]?.autofilledFields?.includes(columnId));
       const value = upload[cellAtDragStart.rowId][columnId];
       if (rowIds.length) {
         dispatch(updateUploadRows(rowIds, { [columnId]: value }));
--- a/src/renderer/state/selection/logics.ts
+++ b/src/renderer/state/selection/logics.ts
@@ -173,10 +173,15 @@ const stopCellDragLogic = createLogic({
   ) => {
     const { cellAtDragStart, rows } = ctx;
     if (cellAtDragStart && rows?.length) {
-      const rowIds = rows.map((row: UploadRowTableId) => row.id);
       const upload = getUpload(getState());
-      const value = upload[cellAtDragStart.rowId][cellAtDragStart.columnId];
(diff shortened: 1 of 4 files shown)


src/renderer/state/selection/logics.ts
```diff
@@ -157,10 +179,18 @@ const stopCellDragLogic = createLogic({
...
     if (cellAtDragStart && rows?.length) {
-      const rowIds = rows.map((row: UploadRowTableId) => row.id);
       const upload = getUpload(getState());
-      const value = upload[cellAtDragStart.rowId][cellAtDragStart.columnId];
-      dispatch(updateUploadRows(rowIds, { [cellAtDragStart.columnId]: value }));
+      const columnId = cellAtDragStart.columnId;
+      // get row ids that dont have autofill
+      const rowIds = rows
+        .map((row: UploadRowTableId) => row.id)
+        .filter((id: any) => !upload[id]?.autofilledFields?.includes(columnId));
```
BrianWhitneyAI, line 187: I think this can be typed to string[]

src/renderer/state/selection/logics.ts
```diff
@@ -98,13 +98,35 @@ const startMassEditLogic = createLogic({
...
+    // user should not be able to mass edit on that row
+    // to avoid potentially overwriting extracted metadata.
+    const selectedRowIds: string[] = action.payload;
+    const upload = getUpload(getState());
+
+    // get set of autofilled fields that we should not mass edit
+    const autofilledFieldSets = selectedRowIds.map(
+      (id) => new Set<string>(upload[id]?.autofilledFields || [])
+    );
+    if (autofilledFieldSets.length) {
```
BrianWhitneyAI, line 120: I think this is redundant

BrianWhitneyAI (approved):

PR #287, opened 2026-05-29
Convert osx paths to vast path

yrkim98:
### Overview
For this ticket https://github.com/aics-int/aics-file-upload-app/issues/284

We traditionally converted osx mount paths in FSS, this causes issues since this conversion needs to happen for MXS as well.

Instead of having this in multiple places on the backend, we decided to move osx mount conversion to the clients- so that paths are normalized to a vast path before entering the backend.

### Changes

src/renderer/util/index.ts: converts paths to vast paths. bit differently than in [fss here](https://github.com/aics-int/file-storage-service/pull/560). Should work even if user mounts VAST not in the standard location in our confluence docs.

src/renderer/state/selection/logics.ts: uses conversion method to convert paths after file selection

src/renderer/util/test/index.test.ts: some tests

### DIFF
--- a/src/renderer/state/selection/test/logics.test.ts
+++ b/src/renderer/state/selection/test/logics.test.ts
@@ -19,6 +19,7 @@ import {
   mockState,
   nonEmptyStateForInitiatingUpload,
 } from "../../test/mocks";
+import { AlertType } from "../../types";
 import { Page } from "../../types";
 import { updateUploadRows } from "../../upload/actions";
 import { UPDATE_UPLOAD_ROWS } from "../../upload/constants";
@@ -126,6 +127,24 @@ describe("Selection logics", () => {
       // after
       await logicMiddleware.whenComplete();
       expect(feedback.selectors.getIsLoading(store.getState())).to.equal(false);
+    });
+
+    it("sets an error alert when a non-VAST path is selected", async () => {
+      const { logicMiddleware, store } = createMockReduxStore(mockState);
+
+      store.dispatch(selections.actions.selectUploadType(UploadType.File));
+      store.dispatch(
+        selections.actions.loadFiles(["/Users/brian/Desktop/not_vast.czi"])
+      );
+
+      await logicMiddleware.whenComplete();
+      const alert = feedback.selectors.getAlert(store.getState());
+      expect(alert).to.not.be.undefined;
+      expect(alert?.type).to.equal(AlertType.ERROR);
+      expect(alert?.message).to.include(
+        "You must select files that are on VAST"
+      );
+      expect(getUpload(store.getState())).to.be.empty;
     });
 
     it("should stop loading on error", async () => {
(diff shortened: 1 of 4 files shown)


BrianWhitneyAI (approved):

src/renderer/util/index.ts
```diff
@@ -35,6 +35,22 @@ const canUserRead = async (filePath: string): Promise<boolean> => {
...
 
+/**
+ * Converts an osx mount path to a VAST path starting with /allen.
+ * removes any prefix before /allen (ex: /Volumes/allen/... -> /allen/...).
+ * Throws an error if /allen is not found in the path.
+ */
+export function convertToVastPath(filePath: string): string {
+  const VAST_PREFIX = "/allen";
+  const allenIndex = filePath.toLowerCase().indexOf(VAST_PREFIX);
+
```
rugeli, line 46: i think this lets paths like `/allen_xyz` slip through, maybe safer to search for `/allen/`? This case came to my mind from personal experience, i name my vast mount `allen_remote` to distinguish it from my local /allen. (But if we don't need to support customized mount names and want users to use the fixed `/allen` mount name, then your approach totally makes sense to me)

yrkim98: youre right! fixed here https://github.com/aics-int/aics-file-upload-app/pull/287/changes/8bad1562a9d52a7a1c2637c792931c66fc40f994

src/renderer/util/index.ts
```diff
@@ -35,6 +35,22 @@ const canUserRead = async (filePath: string): Promise<boolean> => {
   }
 };
 
+/**
+ * Converts an osx mount path to a VAST path starting with /allen.
+ * removes any prefix before /allen (ex: /Volumes/allen/... -> /allen/...).
+ * Throws an error if /allen is not found in the path.
+ */
+export function convertToVastPath(filePath: string): string {
```
hughes036, line 43: What do Windows vast paths look like?  Do they need to be converted?

src/renderer/state/selection/test/logics.test.ts
```diff
@@ -109,7 +110,7 @@ describe("Selection logics", () => {
 
       expect(Object.keys(upload)).to.be.lengthOf(1);
       const file = upload[Object.keys(upload)[0]];
-      expect(file.file).to.equal(FILE_FULL_PATH);
```
hughes036, line 112: What about including a non-happy path test, which exercises your code change?

hughes036 (approved):
Looks ok but

- No handling is needed for Windows paths?  Does node fs just convert them automatically to posix?
- Are there any cases known that should be treated as invalid input (rather than converted)
- See comment about only happy path testing

src/renderer/util/index.ts
yrkim98: windows vast paths do not need to be converted, as they'd start with `/allen`!

src/renderer/state/selection/test/logics.test.ts
yrkim98: added this

PR #292, opened 2026-06-30
Mass edit Imaging Session fix

yrkim98:
### Description
Applying a mass edit that set Plate Barcode + Imaging Session together left Imaging Session blank. The event dispatched for well autofill mistakenly reset Imaging Session whenever a plate barcode was entered and imaging session was being reset.

Additionally, selecting a template with Imaging Session as one of its columns was crashing FUA.

### Files Changed

- src/renderer/state/upload/reducer.ts: When you change the plate barcode, Imaging Session and Well normally get cleared. Now, if the same update also sets a new Imaging Session or Well, that value is kept instead of being wiped. (UPDATE_UPLOAD_ROWS already worked this way)
- src/renderer/state/upload/logics.ts: Make well autofill still fire without dropping any values
- src/renderer/state/upload/test/logics.test.ts: Added test: "preserves imaging session when applied alongside plate barcode"
-  src/renderer/containers/CustomDataTable/selectors.tsx: Skip any template annotation whose name matches a built-in plate column (like Imaging Session), so the same column doesn't get added twice.

### Testing
Tested locally to verify that applying mass edit now works for Imaging Session

### DIFF
--- a/src/renderer/containers/CustomDataTable/selectors.tsx
+++ b/src/renderer/containers/CustomDataTable/selectors.tsx
@@ -159,10 +159,6 @@ const METADATA_COLUMN: CustomColumn = {
 // cause downstream effects like reseting user adjusted
 // column widths for selectors that rely on upload state
 // changes
-// Names of annotations rendered by the dedicated PLATE_RELATED_COLUMNS above.
-// Templates may also include these as regular annotations; rendering both would
-// create duplicate react-table columns (which throws) and break editing for
-// fields like "Imaging Session", so we let the dedicated columns win.
 const PLATE_RELATED_COLUMN_NAMES: ReadonlySet<string> = new Set(
   PLATE_RELATED_COLUMNS.map((column) => column.accessor as string)
 );
--- a/src/renderer/containers/CustomDataTable/selectors.tsx
+++ b/src/renderer/containers/CustomDataTable/selectors.tsx
@@ -159,6 +159,14 @@ const METADATA_COLUMN: CustomColumn = {
 // cause downstream effects like reseting user adjusted
 // column widths for selectors that rely on upload state
 // changes
+// Names of annotations rendered by the dedicated PLATE_RELATED_COLUMNS above.
+// Templates may also include these as regular annotations; rendering both would
+// create duplicate react-table columns (which throws) and break editing for
+// fields like "Imaging Session", so we let the dedicated columns win.
+const PLATE_RELATED_COLUMN_NAMES: ReadonlySet<string> = new Set(
+  PLATE_RELATED_COLUMNS.map((column) => column.accessor as string)
+);
+
 export const getTemplateColumnsForTable = createSelector(
   [getAnnotationTypes, getAppliedTemplate],
   (annotationTypes, template): CustomColumn[] => {
@@ -169,6 +177,9 @@ export const getTemplateColumnsForTable = createSelector(
     return [
       ...PLATE_RELATED_COLUMNS,
       ...template.annotations
+        .filter(
+          (annotation) => !PLATE_RELATED_COLUMN_NAMES.has(annotation.name)
+        )
         .sort((a, b) => a.orderIndex - b.orderIndex)
         .map((annotation) => {
           const type = annotationTypes.find(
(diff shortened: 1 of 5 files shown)


BrianWhitneyAI (approved):

src/renderer/containers/CustomDataTable/selectors.tsx
```diff
@@ -169,6 +173,9 @@ export const getTemplateColumnsForTable = createSelector(
     return [
```
hughes036, line 173: Out of scope, but this return statement badly needs to be refactored.

src/renderer/containers/CustomDataTable/selectors.tsx
```diff
@@ -169,6 +173,9 @@ export const getTemplateColumnsForTable = createSelector(
...
       ...PLATE_RELATED_COLUMNS,
       ...template.annotations
+        .filter(
```
hughes036, line 176: The problem was that without the filter, this aggregation was double counting imaging session right? 

Was this also possible for `barcode`?

It seems to me like there is a root cause to this problem, but this is treating a symptom.

src/renderer/state/upload/logics.ts
```diff
@@ -491,12 +491,25 @@ const updateUploadRowsLogic = createLogic({
...
+        AnnotationName.IMAGING_SESSION
+      ];
+      const well = (metadataUpdate as Partial<FileModel>)[AnnotationName.WELL];
       for (const fileKey of uploadKeys) {
         // dispatch update to plate barcode for each row
         // which will trigger well autofill
         dispatch(
           updateUpload(fileKey, {
             [AnnotationName.PLATE_BARCODE]: [plateBarcode],
+            ...(imagingSession !== undefined && {
```
hughes036, line 507: Looks like you are just extending (to session and well) the same pattern that existed already for `PLATE_BARCODE` ?

src/renderer/state/upload/test/logics.test.ts
```diff
@@ -1248,6 +1248,55 @@ describe("Upload logics", () => {
         [plateBarcode]
       );
     });
+
+    it("preserves imaging session when applied alongside plate barcode", async () => {
```
hughes036, line 1252: If you are just extending the existing special case handling for `PLATE_BARCODE`, is there already an equivalent test for `PLATE_BARCODE`, and could you just add the assert cases for `WELL` and `SESSION` rather than repeat all the test bootstrapping?

hughes036 (approved):

src/renderer/containers/CustomDataTable/selectors.tsx
hughes036: Im worried about all the special case handling we are doing for specific annotations.  To me, this change show that it isn't scale well.

src/renderer/state/upload/test/logics.test.ts
yrkim98: well this is a new special case in which plate barcode AND imaging session is applied together, in which we need to preserve imaging session

src/renderer/state/upload/logics.ts
yrkim98: sort of. The `PLATE_BARCODE` re-dispatch was already there for well autofill. The new addition is forwarding `Imaging_Session` and `well` through it (when they're in the same update) so the reducer's reset doesn't clear them.

src/renderer/containers/CustomDataTable/selectors.tsx
yrkim98: Youre definitely right, there is a way to fix the root cause, but thats a bigger refactor than I'd want to fold into this bugfix though. I'd suggest we ship this bugfix as the stopgap and ill open a ticket for the root cause fix.

yrkim98: https://github.com/aics-int/aics-file-upload-app/issues/294

src/renderer/containers/CustomDataTable/selectors.tsx
yrkim98: i agree, but this has always been here.

should we start a refactor ticket?

yrkim98:
I've created tickets to refactor one thing, but will ship this bugfix out with the two approvals

PR #296, opened 2026-07-27
Imaging session entered now refreshes well

yrkim98:
To fix https://github.com/aics-int/aics-file-upload-app/issues/295

Previously, FUA assumed a plain plate (not associated with an imaging session) always exists for a barcode. Hannah reported this issue with FUA on a barcode which only had plates associated with imaging sessions, which broke the app.

- Autofill could never fire for this barcode. The well autofill code only ran when the update contained a plate barcode.
- Well cell stayed greyed out and did not re-trigger when an imaging session was selected

This PR fixes these issues
`src/renderer/state/upload/logics.ts`: plate lookup + well autofill now triggers on imaging session selection, and mass editor fixed to have the same behavior

`src/renderer/state/upload/test/logics.test.ts` : some tests for this new code

### DIFF
--- a/src/renderer/state/upload/logics.ts
+++ b/src/renderer/state/upload/logics.ts
@@ -365,73 +365,103 @@ const updateUploadLogic = createLogic({
     done: ReduxLogicDoneCb
   ) => {
     const upload = deps.action.payload.upload || deps.action.payload;
+    const fileKey: string | undefined = deps.action.payload.key;
+    const modifiedAnnotations = Object.keys(upload);
 
-    // If a plate barcode is being updated check for imaging sessions
-    const plateBarcode = upload[AnnotationName.PLATE_BARCODE]?.[0];
-    if (plateBarcode) {
-      const plateBarcodeToPlates = getPlateBarcodeToPlates(deps.getState());
-      let updatedPlateBarcodeToPlates = plateBarcodeToPlates;
-      // Avoid re-querying for the imaging sessions if this
-      // plate barcode has been selected before
-      if (!Object.keys(plateBarcodeToPlates).includes(plateBarcode)) {
-        const imagingSessionsForPlateBarcode =
-          await deps.labkeyClient.findImagingSessionsByPlateBarcode(
-            plateBarcode
-          );
-        const imagingSessionsWithPlateInfo: PlateAtImagingSession[] =
-          await Promise.all(
-            imagingSessionsForPlateBarcode.map(async (is) => {
-              const { wells } = await deps.mmsClient.getPlate(
-                plateBarcode,
-                is["ImagingSessionId"]
-              );
+    // Plate wells depend on both the plate barcode and the imaging session
+    // (a barcode may exist only as imaging-session plates, in which case the
+    // wells aren't known until an imaging session is selected), so react to
+    // updates of either
+    if (
+      !modifiedAnnotations.includes(AnnotationName.PLATE_BARCODE) &&
+      !modifiedAnnotations.includes(AnnotationName.IMAGING_SESSION)
+    ) {
+      done();
+      return;
+    }
 
-              return {
-                wells,
-                imagingSessionId: is["ImagingSessionId"],
-                name: is["ImagingSessionId/Name"],
-              };
-            })
-          );
+    // The update may contain only one of plate barcode / imaging session;
+    // fall back to what is already stored for this row (or the mass edit row)
+    const state = deps.getState();
+    const storedRow = fileKey
+      ? getUpload(state)[fileKey]
+      : getMassEditRow(state);
+    const plateBarcode =
+      upload[AnnotationName.PLATE_BARCODE]?.[0] ??
+      storedRow?.[AnnotationName.PLATE_BARCODE]?.[0];
+    if (!plateBarcode) {
+      done();
+      return;
+    }
+    const imagingSessionName =
+      upload[AnnotationName.IMAGING_SESSION]?.[0] ??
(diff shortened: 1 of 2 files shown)


src/renderer/state/upload/logics.ts
```diff
@@ -513,6 +544,21 @@ const updateUploadRowsLogic = createLogic({
           })
         );
       }
+    } else if (imagingSession?.length) {
```
BrianWhitneyAI, line 550: This technically doesn't cover 'imagingSession = [] '

src/renderer/state/upload/logics.ts
```diff
@@ -365,73 +365,103 @@ const updateUploadLogic = createLogic({
...
+            };
+          })
+        );
 
-        updatedPlateBarcodeToPlates = {
-          ...plateBarcodeToPlates,
-          [plateBarcode]: imagingSessionsWithPlateInfo,
-        };
-        dispatch(setPlateBarcodeToPlates(updatedPlateBarcodeToPlates));
+      // If the barcode has no imaging sessions, find info of plate without
```
BrianWhitneyAI: maybe say how?

BrianWhitneyAI (approved):

src/renderer/state/upload/logics.ts
```diff
@@ -365,73 +365,103 @@ const updateUploadLogic = createLogic({
...
+      if (!imagingSessionsWithPlateInfo.length) {
+        const { wells } = await deps.mmsClient.getPlate(plateBarcode);
+        imagingSessionsWithPlateInfo.push({ wells });
```
saeliddp: This reads a bit strange to me. We've failed to find `imagingSessionsWithPlateInfo`, and now we're pushing wells (plate info), detached from any imaging session, into `imagingSessionsWithPlateInfo`?

src/renderer/state/upload/logics.ts
```diff
@@ -365,73 +365,103 @@ const updateUploadLogic = createLogic({
...
+    const colValue = mxsData?.["Column"]?.value;
+
+    if (fileKey && rowValue !== undefined && colValue !== undefined) {
+      // row and col are 1-indexed in metadata, convert to 0-indexed
+      const row = Number(rowValue) - 1;
+      const col = Number(colValue) - 1;
+      const plates = updatedPlateBarcodeToPlates[plateBarcode];
+      // this should match the logic in WellCell
+      const plate = plates?.find((p) =>
+        imagingSessionName ? p.name === imagingSessionName : !p.name
```
saeliddp, line 458: I could use some help understanding this bit. If we don't know `imagingSessionName`, then the filter criteria becomes `!p.name`--does a falsy `p.name` mean something special??

src/renderer/state/upload/logics.ts
```diff
@@ -490,14 +520,15 @@ const updateUploadRowsLogic = createLogic({
       AnnotationName.PLATE_BARCODE
     ]?.[0]; // update contains plate barcode value, or undefined if no plateBarcode value
 
+    // Preserve any imaging session / well provided in the same update (e.g.
+    // a mass edit) so the plate barcode re-dispatch below does not wipe them
+    // out via the UPDATE_UPLOAD reset logic.
+    const imagingSession = (metadataUpdate as Partial<FileModel>)[
+      AnnotationName.IMAGING_SESSION
+    ];
+    const well = (metadataUpdate as Partial<FileModel>)[AnnotationName.WELL];
```
saeliddp, line 532: So this allows a user to override the auto-selected well?

saeliddp (commented):
I left a few questions. Holding off on approval until I better understand these changes

src/renderer/state/upload/logics.ts
yrkim98: this is preexisting code that the diff is showing wrong (see above)

it is correct (albeit a bit misleading with the variable name). it's a list of plates for a barcode, each plate is somtimes tied to an imaging session- no imaging session is normal for 99% of barcode and the plate without an imaging session still exists and can be fetched and pushed

although it is existing code i have renamed this to `platesForBarcode` which is more indicative of what it actually is

src/renderer/state/upload/logics.ts
yrkim98: `p.name` is just the imaging session name for the plate, it's only set if that plate is tied to an imaging session.

the plate w/o an imaging session (the _default_ plate) has no name at all.

so if `imagingSessionName` is undefined (user hasnt selected an imaging session) `!p.name` finds the default plate, and if it is defined we pick the plate to match `imagingSessionName`

src/renderer/state/upload/logics.ts
yrkim98: this is again existing code, dunno why the diff is all funky

but the code is explained in this pr https://github.com/aics-int/aics-file-upload-app/pull/292

src/renderer/state/upload/logics.ts
yrkim98: thats intentional, [] means the session is being cleared so theres nothing to autofill against

src/renderer/state/upload/logics.ts
saeliddp: gotcha, so it's guaranteed that there's only ever one plate for a given barcode with `!p.name = true`?

yrkim98: correct!

saeliddp (approved):
Approving with ~80% confidence due to my limited FUA knowledge. (and assuming you've manually tested and confirmed this fix on real data)

PR #298, opened 2026-08-11
Fix sync abandoned upload routine

yrkim98:
The old `syncAbandonedUploads()` routine had a error (when JSS is succeeded it only returned true() without running complete().

This simplifies the upload routine and adds that

### DIFF
--- a/src/renderer/services/file-management-system/index.ts
+++ b/src/renderer/services/file-management-system/index.ts
@@ -198,34 +198,17 @@ export default class FileManagementSystem {
    * while app is closed. Used on app restart to update any abandoned uploads.
    */
   public async syncAbandonedUploadStatus(uploadId: string): Promise<boolean> {
-    const fuaUpload = (await this.jss.getJob(uploadId)) as UploadJob;
+    const fssManagedJSSJob = (await this.jss.getJob(uploadId)) as UploadJob;
+    const fssStatus = await this.fss.getStatus(uploadId);
 
-    // checking jss first to see if job is complete there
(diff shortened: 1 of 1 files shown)


hughes036 (approved):
nit: could you look into unit testing this?  seems useful and feasible using mocks.

src/renderer/services/file-management-system/index.ts
```diff
@@ -198,34 +198,17 @@ export default class FileManagementSystem {
    * while app is closed. Used on app restart to update any abandoned uploads.
    */
   public async syncAbandonedUploadStatus(uploadId: string): Promise<boolean> {
-    const fuaUpload = (await this.jss.getJob(uploadId)) as UploadJob;
-
-    // checking jss first to see if job is complete there
-    if (fuaUpload.status === JSSJobStatus.SUCCEEDED) {
+    const fssManagedJSSJob = (await this.jss.getJob(uploadId)) as UploadJob;
+    const fssStatus = await this.fss.getStatus(uploadId);
```
BrianWhitneyAI, line 202: Would there be a way to go off one request?

BrianWhitneyAI (approved):

src/renderer/services/file-management-system/index.ts
BrianWhitneyAI: Like does the JSS job status sync up with the FSS status?

yrkim98: I dont think so, we'd need to check both as only one might be reporting a complete staus

PR #299, opened 2026-08-12
Loading wheel for mass editor and drag and drop applies

yrkim98:
**Problem**
When a user does a mass editor or drag and drop apply for many files, it looks like theres a short freeze while the edit is written into every row and the table renders.

**My solution**
I added a loading wheel that appears right after the edits are applied, and the table re renders happen. the loading wheel stays up through the UI blocking part, and goes down once the renders finish.

This loading wheel finishes before the metadata extraction plate barcode lookups happen per file, but that already has a loading indicator for each row and the UI is not frozen during this so I thought this would be the best solution.

used this as a resource: https://medium.com/@owencm/one-weird-trick-to-performant-touch-response-animations-with-react-9fe4a0838116

Screenshot 2026-08-12 at 8 05 18 AM

**Changes**
src/renderer/state/selection/logics.ts: `startLoadingAndAwaitPaint` is the most important thing here. It shows the UI spinner, waits for that to render before returning so we're guaranteed to show the loading wheel before other callers block the UI thread. [Used the above medium link as a resource for this](https://medium.com/@owencm/one-weird-trick-to-performant-touch-response-animations-with-react-9fe4a0838116)

src/renderer/state/upload/logics.ts: dispatches an event that stops the loading wheel 

src/renderer/containers/AddMetadataPage/index.tsx: shows the loading wheel on the metadata page

src/renderer/containers/AddMetadataPage/styles.pcss: style for the loading wheel (claude helped me with this)

src/renderer/state/selection/test/logics.test.ts: some tests from claude

### DIFF
--- a/src/renderer/state/selection/logics.ts
+++ b/src/renderer/state/selection/logics.ts
@@ -122,7 +122,7 @@ const startMassEditLogic = createLogic({
 });
 
 const applyMassEditLogic = createLogic({
-  process: (
+  process: async (
     { ctx }: ReduxLogicProcessDependencies,
     dispatch: ReduxLogicNextCb,
     done: ReduxLogicDoneCb
@@ -137,6 +137,15 @@ const applyMassEditLogic = createLogic({
         }),
       }),
       {}
+    );
+    // mass edit blocks the UI thread while metadata is extracted for all the files
+    // so we  show a loading indicator and wait for it to render before starting
+    // updateUploadRowsLogic will later stop the indicator when the changes are done
+    dispatch(startLoading());
+    await new Promise((resolve) =>
+      typeof requestAnimationFrame === "function"
+        ? requestAnimationFrame(() => requestAnimationFrame(resolve))
+        : setTimeout(resolve, 0)
     );
     dispatch(updateUploadRows(rowIds, rowData));
     done();
--- a/src/renderer/state/selection/logics.ts
+++ b/src/renderer/state/selection/logics.ts
@@ -121,6 +121,20 @@ const startMassEditLogic = createLogic({
   type: START_MASS_EDIT,
 });
 
+/**
+ * shows a loading indicator which spins while the caller blocks the UI thread
+ * used for mass edit applies and drag and drop applies across many files
+ * the caller need to call stopLoading once the updates are applied
+ */
+async function startLoadingAndAwaitPaint(dispatch: ReduxLogicNextCb) {
+  dispatch(startLoading());
+  await new Promise((resolve) =>
+    typeof requestAnimationFrame === "function"
+      ? requestAnimationFrame(() => requestAnimationFrame(resolve))
+      : setTimeout(resolve, 0)
+  );
+}
+
 const applyMassEditLogic = createLogic({
   process: async (
     { ctx }: ReduxLogicProcessDependencies,
@@ -138,15 +152,9 @@ const applyMassEditLogic = createLogic({
       }),
       {}
     );
-    // mass edit blocks the UI thread while metadata is extracted for all the files
-    // so we  show a loading indicator and wait for it to render before starting
-    // updateUploadRowsLogic will later stop the indicator when the changes are done
-    dispatch(startLoading());
-    await new Promise((resolve) =>
-      typeof requestAnimationFrame === "function"
-        ? requestAnimationFrame(() => requestAnimationFrame(resolve))
-        : setTimeout(resolve, 0)
-    );
+    // mass edit blocks the UI thread while every row is updated, show loading wheel
+    // updateUploadRowsLogic stops the indicator once the changes are done
+    await startLoadingAndAwaitPaint(dispatch);
     dispatch(updateUploadRows(rowIds, rowData));
     done();
   },
@@ -164,7 +172,7 @@ const applyMassEditLogic = createLogic({
 });
 
 const stopCellDragLogic = createLogic({
-  process: (
+  process: async (
     { ctx, getState }: ReduxLogicProcessDependencies,
     dispatch: ReduxLogicNextCb,
     done: ReduxLogicDoneCb
@@ -183,6 +191,9 @@ const stopCellDragLogic = createLogic({
 
       // drag to copy protection against overwriting rows already autofilled by mxs
       if (notAutofilledRowIds.length) {
+        // drag to copy blocks the UI thread just like a mass edi
+        // updateUploadRowsLogic stops the indicator once the changes are done
+        await startLoadingAndAwaitPaint(dispatch);
         dispatch(updateUploadRows(notAutofilledRowIds, { [columnId]: value }));
       }
     }
--- a/src/renderer/state/selection/test/logics.test.ts
+++ b/src/renderer/state/selection/test/logics.test.ts
@@ -377,6 +377,42 @@ describe("Selection logics", () => {
 
       // Assert
       expect(actions.includesType(UPDATE_UPLOAD_ROWS)).to.be.false;
+      // No rows to update, so the loading indicator never shows
+      expect(actions.includesType(feedback.actions.startLoading().type)).to.be
+        .false;
+    });
+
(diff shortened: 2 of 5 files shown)


src/renderer/state/selection/logics.ts
```diff
@@ -121,8 +121,22 @@ const startMassEditLogic = createLogic({
   type: START_MASS_EDIT,
 });
 
+/**
+ * shows a loading indicator which spins while the caller blocks the UI thread
+ * used for mass edit applies and drag and drop applies across many files
+ * the caller need to call stopLoading once the updates are applied
+ */
+async function startLoadingAndAwaitPaint(dispatch: ReduxLogicNextCb) {
```
BrianWhitneyAI, line 129: Paint?

BrianWhitneyAI (approved):

src/renderer/state/selection/logics.ts
```diff
@@ -121,8 +121,22 @@ const startMassEditLogic = createLogic({
   type: START_MASS_EDIT,
 });
 
+/**
+ * shows a loading indicator which spins while the caller blocks the UI thread
```
hughes036, line 125: What specifically is happening when the caller blocks the UI?  

Is it fetch calls to MMS?  fetch is something synchronous blocking the event loop?

When you say main thread, I think what you actually mean is the Event Loop.  Similar in effect but an import distinction in concept.

src/renderer/state/selection/logics.ts
```diff
@@ -138,6 +152,9 @@ const applyMassEditLogic = createLogic({
       }),
       {}
     );
+    // mass edit blocks the UI thread while every row is updated, show loading wheel
```
hughes036, line 155: what is happening while the UI is blocked?  Compute?  io?

src/renderer/state/selection/logics.ts
```diff
@@ -121,8 +121,22 @@ const startMassEditLogic = createLogic({
...
 
+/**
+ * shows a loading indicator which spins while the caller blocks the UI thread
+ * used for mass edit applies and drag and drop applies across many files
+ * the caller need to call stopLoading once the updates are applied
+ */
+async function startLoadingAndAwaitPaint(dispatch: ReduxLogicNextCb) {
+  dispatch(startLoading());
+  await new Promise((resolve) =>
+    typeof requestAnimationFrame === "function"
```
hughes036, line 132: Can you document what is happening in this promise?  Im a bit lost.

src/renderer/state/selection/logics.ts
```diff
@@ -121,8 +121,22 @@ const startMassEditLogic = createLogic({
...
+      ? requestAnimationFrame(() => requestAnimationFrame(resolve))
+      : setTimeout(resolve, 0)
```
hughes036, line 134: A delay of 0 ?

src/renderer/state/selection/test/logics.test.ts
```diff
@@ -218,6 +218,28 @@ describe("Selection logics", () => {
...
+
+      // Act
+      store.dispatch(applyMassEdit());
+      await logicMiddleware.whenComplete();
+
+      // Assert
+      expect(actions.includesType(feedback.actions.startLoading().type)).to.be
+        .true;
+      // updateUploadRowsLogic stops the indicator once the rows are updated
+      expect(feedback.selectors.getIsLoading(store.getState())).to.be.false;
```
hughes036, line 241: The first assert is that the wheel should be displayed, and the second is that it should not-right?
why do they use different methods to check if the loading wheel is displayed?

what changes in between the 2 asserts here?

src/renderer/state/selection/test/logics.test.ts
```diff
@@ -355,6 +377,42 @@ describe("Selection logics", () => {
...
+
+      // Act
+      store.dispatch(stopCellDrag());
+      await logicMiddleware.whenComplete();
+
+      // Assert
+      expect(actions.includesType(feedback.actions.startLoading().type)).to.be
+        .true;
+      // updateUploadRowsLogic stops the indicator once the rows are updated
+      expect(feedback.selectors.getIsLoading(store.getState())).to.be.false;
```
hughes036, line 415: again, im not understanding what happened to in between the 2 asserts to disable the wheel 

If the initial action was `store.dispatch(_start_CellDrag());`
and If it invoked `store.dispatch(stopCellDrag());` before the second assert, that would make sense to me.

src/renderer/state/selection/test/logics.test.ts
```diff
@@ -355,6 +377,42 @@ describe("Selection logics", () => {
...
+        upload: getMockStateWithHistory({
+          [cellAtDragStart.rowId]: {
+            file: "/some/path/to/a/file.txt",
+            [cellAtDragStart.columnId]: "false",
+          },
+        }),
+      });
+
+      // Act
+      store.dispatch(stopCellDrag());
```
hughes036, line 408: should the action be _start_CellDrag?

src/renderer/state/selection/logics.ts
yrkim98: see https://medium.com/@owencm/one-weird-trick-to-performant-touch-response-animations-with-react-9fe4a0838116

src/renderer/state/selection/logics.ts
yrkim98: yeah its just synchronous javascript. the fetch calls to mms already happen async and have a separate loading wheel.

so when edit is applied, redux dispatches an update per row, and each dispatch causes react to rerender the table, which causes a huge slowdown for changes to many rows

src/renderer/state/selection/logics.ts
yrkim98: just compute, see https://github.com/aics-int/aics-file-upload-app/pull/299#discussion_r3799448079

src/renderer/state/selection/logics.ts
yrkim98: added comments, but for more info see the medium thread in the description

src/renderer/state/selection/logics.ts
yrkim98: its needed for unit tests (was failing before, claude  suggested add for unit tests to pass)

src/renderer/state/selection/test/logics.test.ts
yrkim98: nothing changes since the assert runs after everything completes, but the first assert checks that the mock started loading a wheel at some point (so a wheel was shown at some point)
second assert checks that the wheel is off after everything ran (so the wheel is now not shown, and is off)

src/renderer/state/selection/test/logics.test.ts
yrkim98: see https://github.com/aics-int/aics-file-upload-app/pull/299#discussion_r3799912471

src/renderer/state/selection/test/logics.test.ts
yrkim98: we are testing the `stopCellDragLogic()` and `stopCellDrag()` is the right dispatch here. (that simulates mouse release)

src/renderer/state/selection/logics.ts
saeliddp: Have you looked into options to solve the root cause of the slowness? I imagine that change would be more involved, but it could be good to make a ticket for it if there are possibilities.

saeliddp (approved):
Approving with limited understanding of the underlying code since it looks like you were able to manually test this. Would recommend a review from someone with more domain knowledge, if possible!

src/renderer/state/selection/logics.ts
hughes036: Can you explain why it was failing, and what claude's suggestion fixed ?

src/renderer/state/selection/logics.ts
hughes036: It seems like a better solution is to only re-render after the MMS reqs have all been recieved.

